PR #23, opened 2015-06-11
Add a make file for aes

abcjjy:
Fix a bug in minizip's command line option parsing code.
Fix a issue in aes when building on unix like system
Add marco to use constant salt to save unzip time.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,10 @@
+*.o
+libaes.a
+libz.a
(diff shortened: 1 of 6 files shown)


nmoinvaz:
I have merged some of the changes. What was the bug in the command line parsing code?

nmoinvaz:
I have also added in the command line parsing fix.

PR #24, opened 2015-06-15
Fix use-of-uninitialized-value warning.

LindleyF:
UNZ_EOF and UNZ_OK are both zero, so the caller of unzlocal_getByte has no way to know whether or not pi points to data or garbage. This change makes it point to zero in the EOF case. While it might be better to indicate an error immediately, that would be a more significant change. The library ought to be just as good at handling an erroneous zero byte as anything else; and this way dynamic analysis tools won't complain about use-of-uninitialized-value when reading a malformed file.


LindleyF:
Ping?

nmoinvaz:
Looks good. Thank you!

PR #34, opened 2015-09-10
Honor "version made by" set by user. Allow to set "version made by" in Zip64 end of central directory.

juanii:
I found that if you use zipOpenNewFileInZip4_64 function to open a new file within a ZIP archive with a specific versionMadeBy value, whenever the entry needs to be converted to Zip64 the versionMadeBy value set by the user was overwritten with the value 45 (0x002d). This patch honors the previous field's high byte value (see 4.4.2.1 and 4.4.2.2 in APPNOTE.TXT) and doesn't change the low byte value if it's already greater than 45.

Also, new version of zipClose, zipClose2_64 was created to allow the user to set the "version made by" field in the Zip64 end of central directory record. Another function zipClose_64 was created but it's there just to fill in the gap (tried to follow some convention I think was used in other functions like zipOpenNewFileInZip*).


juanii:
I've just fixed some conflicts with commits made after the pull request. Are there any other issues preventing this request from being accepted? I'd be glad to fix them.

nmoinvaz:
I have incorporated it now. Thanks for the contribution!

PR #38, opened 2015-10-17
Fixed an arithmetic overflow leading to unzipping failure in 64bit mode

franiu:
It seems there's quite a few local variables and struct fields, which overflow when file sizes go above 4GB (32bit boundary). Most of the overflows seem to be benign, but one in particular causes the pfile_in_zip_read_info->total_out_64 to go back to 0, which results in reading past the file compressed buffer.

Surprisingly this issue only surfaces in 64-bit builds. My theory (which I didn't fully verify) is that the stream and bstream fields overflow in sync in 32-bit environment. I've patched it and it seems to be working fine in my setup (iOS project, verified with huge archives on i386/x86_64 sim and armv7/arm64 devices, llvm compiler), but I don't have the resources to verify/patch the problem for other platforms/compilers.

### DIFF
--- a/unzip.c
+++ b/unzip.c
@@ -1349,10 +1349,10 @@ extern int ZEXPORT unzReadCurrentFile(unzFile file, voidp buf, unsigned len)
     {
         if (pfile_in_zip_read_info->stream.avail_in == 0)
         {
-            uInt bytes_to_read = UNZ_BUFSIZE;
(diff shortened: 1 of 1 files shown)


nmoinvaz:
Thank you!

franiu:
No problem :-)

PR #53, opened 2016-03-25
Fix Build Error

matthew1471:
Build fails when NOUNCRYPT is set.


nmoinvaz:
It looks like you are using tabs instead of spaces. Can you please fix that? Thanks!

matthew1471:
Apologies, Visual Studio 2015 defaults :-).

nmoinvaz:
Thank you!!

PR #54, opened 2016-04-02
added cmake support for shared build and fixed some typos. Added expo…

mschmieder:
Hey,
just worked a little bit more on the CMake integration. Found out that there were still some errors in it. Added export target definitions, correct install directories and also added shared build support

- Matthias


nmoinvaz:
Thanks!

PR #56, opened 2016-04-09
Added ability to specify ZLIB_ROOT when using CMAKE

EosPengwern:
I'm experimenting with different versions of zlib (Intel/Kukunas, CloudFlare etc.) and I found it useful to be able to link minizip with a particular version out of those that I have available. I've modified the CMakeLists.txt to make this straightforward to do.


nmoinvaz:
It looks like there are conflicts. Can you fix them?

EosPengwern:
I'll have a go, but probably won't have the opportunity until tomorrow.

EosPengwern:
Actually, that was more straightforward than I'd hoped. I've done it already.

PR #57, opened 2016-04-16
Moved some higher-level subrotuines into zip.c and unzip.c

EosPengwern:
I found that, by moving some higher-level routines such as check_file_exists and do_extract_all from the minizip.c and miniunz.c executables into the zip.c and unzip.c files within the static library, I was able to significantly reduce the amount of code I had to copy and paste into my own application without having to change the behaviour or interfaces of the minizip library itself.This is the version I'm now using.

### DIFF
--- a/miniunz.c
+++ b/miniunz.c
@@ -14,163 +14,25 @@
    See the accompanying LICENSE file for the full text of the license.
 */
 
-#if (!defined(_WIN32)) && (!defined(WIN32)) && (!defined(__APPLE__))
-#  ifndef __USE_FILE_OFFSET64
-#    define __USE_FILE_OFFSET64
-#  endif
-#  ifndef __USE_LARGEFILE64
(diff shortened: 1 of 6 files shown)


nmoinvaz:
I have thought about doing this type of thing as well, but I think I will leave the code as it is. Some people don't ever use minizip.c or miniunz.c. Some people use zip.c only or unzip.c only, which is why there is some duplication between in those two files.

EosPengwern:
That's fine; I'll leave the modified version in my own account for anyone else who may want it, but I can see the rationale for keeping the API of your version 'pristine'.

nmoinvaz:
I think it might be better to put the shared code that is used in miniunz.c and minizip.c into another file, other than zip.c because a lot of the code is platform specific.

nmoinvaz:
Made changes in dev branch. Thanks.

PR #59, opened 2016-05-30
Fixed opening for read while zip locked by opening to write.

sergey-shambir:
It's possible to read file even if opened to write by the same process.
To enable this feature, I've added FILE_SHARE_WRITE flag.
I've also added flagsAndAttributes initialization by more appropriate
constant 0x80.


nmoinvaz:
Thanks this has been resolved in the dev branch.

PR #60, opened 2016-06-08
Code Review: Replace ZipKit (#9508)

opsGavin:
Fixed a couple of compiler issues on OSX builds.

### DIFF
--- a/ioapi.c
+++ b/ioapi.c
@@ -107,7 +107,7 @@ static int     ZCALLBACK ferror_file_func OF((voidpf opaque, voidpf stream));
 typedef struct 
 {
     FILE *file;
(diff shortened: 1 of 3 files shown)


nmoinvaz:
Can you remove the changes to the readme from the pull request. Thanks!

opsGavin:
Sorry @nmoinvaz - I actually meant this PR to be against a private fork of minizip. I guess that's the danger with relying on automated scripts...

I'm more than happy to push these OSX warning fixes back to the original, but I should say that I haven't tested building with Visual Studio and we're also using the `NOCRYPT` path so there may be some other warnings which I've not dealt with.

I've removed the changes to the readme though.

PR #63, opened 2016-07-29
Check password verification bytes.

jlawton:
I believe that the password verification bytes were intended to be checked.

I added a new error value, because I think the distinction would be useful in practice to handle the error.


nmoinvaz:
Thanks!

PR #65, opened 2016-08-16
Changed compiler flag "HAVE_AES" to be platform independent

TermeHansen:
Changed compiler flag "HAVE_AES" to be set by platform independent add_definitions
Added "-D_CRT_SECURE_NO_DEPRECATE" to WIN32 compiler and fPIC to gcc compiler


nmoinvaz:
Thanks!

PR #78, opened 2016-11-07
Fix error C2664 on MSVC 2015

inequation:
I'm compiling minizip in my project by #including its source code inline (namely, `unzip.c` and `ioapi.c`) in MSVC 2015. I'll let the compiler speak for itself:
```
ioapi.c(201): error C2664: 'char *strncpy(char *,const char *,std::size_t)': cannot convert argument 2 from 'void *' to 'const char *'
ioapi.c(201): note: Conversion from 'void*' to pointer to non-'void' requires an explicit cast
```
There are two more in the same vein. This PR is a trivial fix for these errors.

### DIFF
--- a/ioapi.c
+++ b/ioapi.c
@@ -120,7 +120,7 @@ static voidpf file_build_ioposix(FILE *file, const char *filename)
     ioposix->file = file;
     ioposix->filenameLength = (int)strlen(filename) + 1;
     ioposix->filename = (char*)malloc(ioposix->filenameLength * sizeof(char));
-    strncpy(ioposix->filename, filename, ioposix->filenameLength);
+    strncpy((char*)ioposix->filename, filename, ioposix->filenameLength);
     return (voidpf)ioposix;
 }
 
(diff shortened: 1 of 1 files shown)


MacGritsch:
Why not change "void *filename" to "char *filename" in the struct?

inequation:
Yes, that would probably be better still. I've simply chosen to do the casting because the surrounding code does the same, including the API (cf. lines 122-123, 146 etc.).

I can't see any reason to use a `void` pointer in the first place, though. Perhaps this was in anticipation of also supporting wide characters, or using miscellaneous data structure as "filename" (i.e. not necessarily strings), but then using regular `char` string functions probably defeats the purpose.

MacGritsch:
I saw the lines above your changes and there the malloc-result is casted to (char *) so I think it was changed later to void...

inequation:
Digging through history, it's been like this since January 2012 (commit 6d35b6b8845fae3d3cbde67bdad40e47e4623cf8; also, GitHub is very shitty at displaying this diff for whatever reason, I had to use an external tool to view it 😕). Apparently, this data structure is used for generating part-numbered filenames for archives spanning multiple disks.

I'm all for changing the field type instead, but I'd like a project maintainer to weigh in first. :)

nmoinvaz:
Change looks good.

inequation:
@nmoinvaz Great! Can we expect it to be merged?

inequation:
Thank you!

PR #84, opened 2017-01-28
Value stored is never read

Coeur:
Xcode is giving warnings for unused values.

(maybe this pull request doesn't need to be merged as is, but it gives a hint to where we may have forgot to handle something)

### DIFF
--- a/ioapi_buf.c
+++ b/ioapi_buf.c
@@ -355,7 +355,7 @@ int fseek_buf_internal_func(voidpf opaque, voidpf stream, uint64_t offset, int o
                     streamio->writebuf_pos += (uint32_t)offset;
                     return 0;
                 }
-                offset -= (streamio->writebuf_len - streamio->writebuf_pos);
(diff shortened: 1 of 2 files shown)


Coeur:
rebased PR to latest

nmoinvaz:
I think a few of these have been gotten rid of.

nmoinvaz:
Please rebase PR to latest.

Coeur:
@nmoinvaz rebase done.

What I would merge for sure:
* newline encoding change in `free_datablock` in zip.c
* adding some `if (err == ZIP_OK)` in `zipOpenNewFileInZip4_64` in zip.c

What is symptomatic of something we may not be doing correctly and that may require better fix:
* offset handling in `fseek_buf_internal_func` in ioapi_buf.c
* error handling of `zipFlushWriteBuffer(zi)` in `zipCloseFileInZipRaw64` in zip.c

nmoinvaz:
Thanks!

PR #85, opened 2017-01-28
fix Dereference of null pointer if flagBase & 1 and password is NULL

Coeur:
Xcode is giving warning for `*passwd` when password is NULL.

### DIFF
--- a/zip.c
+++ b/zip.c
@@ -1054,6 +1054,10 @@ extern int ZEXPORT zipOpenNewFileInZip4_64(zipFile file, const char* filename, c
         zi->ci.method = AES_METHOD;
 #endif
     }
+    else
+    {
+        zi->ci.flag &= ~1;
+    }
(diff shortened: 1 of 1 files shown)


nmoinvaz:
I have merged in your changes manually. Thanks!

Coeur:
The reason I changed it to `password != NULL` is for the 21520 developers using [SSZipArchive](https://cocoapods.org/pods/SSZipArchive) who are going to update to Xcode 8.3 soon. They will all see this by default:
screen shot 2017-03-03 at 14 25 53

nmoinvaz:
Does that remove all those warnings or just the last one?

Coeur:
There is only one warning in the screen shot :)
(it's a logic trace)

nmoinvaz:
Fixed manually. Thanks again.

PR #88, opened 2017-02-07
fix endless loop in zipFlushWriteBuffer() 

zett42:
Loop control variable 'write' never changed because 'written' was no longer assigned.
This causes the ZIP file to grow until disk is full.


nmoinvaz:
Thanks!

PR #112, opened 2017-06-05
fix Memory Error with AES

### LINKED ISSUE #111: "Memory Error" warning when using AES
This is an Xcode 8 buildtime warning.
screen shot 2017-06-05 at 14 03 21

It says that memory allocated on `pfile_in_zip_read_info = (file_in_zip64_read_info_s*)ALLOC(sizeof(file_in_zip64_read_info_s));` is not freed in case `password == NULL`.

Coeur:
Fixes #111


nmoinvaz:
Nice find. Thanks!

PR #116, opened 2017-06-09
including minishared as an optional dependency

Coeur:
SSZipArchive is using minishared for its `tm_to_dosdate()`, so it's best to include minishared as a subspec.


Coeur:
By the way, @nmoinvaz I would add a new release tag like **1.2** if I were you.

nmoinvaz:
I will consider the tag. Thanks!

PR #124, opened 2017-07-24
avoid unnecessary cast

Coeur:
pcrc_32_tab is `const z_crc_t *` (aka 'const unsigned long *')
get_crc_table is `const z_crc_t FAR *` (aka 'const unsigned long *')

And clang is unhappy with the previous cast:
>/minizip/zip.c:1235:32: Incompatible pointer types assigning to 'const z_crc_t *' (aka 'const unsigned long *') from 'const unsigned int *'


Coeur:
If there is a need to downcast, then the syntax may be a double cast:

    zi->ci.pcrc_32_tab = (const z_crc_t *)(const unsigned int *)get_crc_table();

PR #125, opened 2017-07-25
updating minizip info

Coeur:


Coeur:
There is also some other strings that are obsolete, but not sure what to do:

in zip.c:

    const char zip_copyright[] = " zip 1.01 Copyright 1998-2004 Gilles Vollant - http://www.winimage.com/zLibDll";

in unzip.c:

    const char unz_copyright[] = " unzip 1.01 Copyright 1998-2004 Gilles Vollant - http://www.winimage.com/zLibDll";

nmoinvaz:
I think it would be better to remove the printf that has the URL in it.

Coeur:
I prefer to keep a reference as an URL, especially with the code license which may require to keep track of the contributors.

But if you do remove any URL, please make sure to update the version numbers to clear distinct values from winimage values. And prefer semantic versioning, like zip 1.2.0 and unzip 1.2.0. It's a good opportunity to also update our old "1.1" version number.

nmoinvaz:
I see what you mean. Then updating all the version numbers and URLs seems like the best way to go.

Coeur:
Done, with most recent copyright on top for clarity.

nmoinvaz:
Thank you!

PR #126, opened 2017-07-25
__attribute__((__unused__)) for crc_for_crypting

Coeur:
If we're using the File Time, then crc_for_crypting becomes __unused__


Coeur:
Eventually, for non-GNU compiler we could keep a:

```
#ifndef __GNUC__
    (crc_for_crypting);
#endif 
```

PR #132, opened 2017-08-24
Clear the stat struct properly to avoid compiler warning

asmaloney:
The original code only initializes the first field which causes clang to spit out "missing field 'st_mode' initializer".

### DIFF
--- a/minishared.c
+++ b/minishared.c
@@ -43,9 +43,11 @@ uint32_t get_file_date(const char *path, uint32_t *dos_date)
         ret = 1;
     }
 #else
-    struct stat s = { 0 };
+    struct stat s;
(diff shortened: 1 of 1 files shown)


Coeur:
You're maybe correct. Just, by curiosity, would clang still complain if we declare as follow?

    struct stat s = { 0, 0, 0, 0, 0, 0, 0 };

asmaloney:
The size & layout of `struct stat` is/may be different on different platforms, so trying to initialize it that way isn't going to work. (For example, the warning just points at a different field now - 'st_atimespec'.)

> This would initialize st_mode without using memset.

Not sure why this should be a goal? Nothing wrong with memset - this is exactly what it's for!

PR #136, opened 2017-09-12
fix Unknown type name 'z_crc_t'

Coeur:
The issue is affecting users with [zlib 1.2.5](https://opensource.apple.com/source/zlib/zlib-61.20.1/) (available on Xcode 7.x), as it doesn't define z_crc_t yet.

This workaround is not needed anymore with [zlib 1.2.7](https://sourceforge.net/p/miranda/svn/14263/tree/trunk/miranda/plugins/zlib/) (1.2.8 is available on Xcode 8.3) as it defines z_crc_t in zconf.h.

### DIFF
--- a/crypt.h
+++ b/crypt.h
@@ -16,6 +16,10 @@
 #ifndef _MINICRYPT_H
 #define _MINICRYPT_H
 
+#if ZLIB_VERNUM < 0x1270
+typedef unsigned long z_crc_t;
+#endif
+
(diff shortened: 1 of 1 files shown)


nmoinvaz:
Thanks!

jhudsonWA:
@Coeur @nmoinvaz Seems like this change was removed from version 2? Getting the same error in Xcode 7 in the file: mz_strm_crypt.c

PR #139, opened 2017-09-17
fix miniunz files date for non-WIN32 and adding helper dosdate_to_time_t

Coeur:
In [minishared L137](https://github.com/nmoinvaz/minizip/blob/master/minishared.c#L137) we got this comment:
> (assumed to be between 1980 and 2107, typical output of old software that does 'year-1900' to get a double digit year)

Well, the 'year-1900' is not just for old software, it's because it's defined as a standard in ISO/IEC 9899: https://www.cs.helsinki.fi/group/boi2016/doc/cppreference/reference/en.cppreference.com/w/c/chrono/tm.html.

So when we're adding `+ 1980` on  [minishared L108](https://github.com/nmoinvaz/minizip/blob/master/minishared.c#L108), we're breaking a standard. We should instead be adding `+ 80` only (+ 1980 - 1900). Well, it's a bit late to fix it, so instead I'd like to introduce a helper to convert both a _standard_ `tm` and a _minizip_ `tm` to a `time_t` using `mktime`. **In fact, `mktime` only works correctly with standard `tm`**, so we need to subtract 1900 when dealing with a _minizip_ `tm`, and to do so I'm using `% 1900`.

Also adding a bit of refactoring because `mkdir` does not require to validate the ranges of `struct tm` components.

### DIFF
--- a/minishared.c
+++ b/minishared.c
@@ -99,7 +99,8 @@ void change_file_date(const char *path, uint32_t dos_date)
 #endif
 }
 
-int dosdate_to_tm(uint64_t dos_date, struct tm *ptm)
+// convertion without validation
+void dosdate_to_raw_tm(uint64_t dos_date, struct tm *ptm)
 {
     uint64_t date = (uint64_t)(dos_date >> 16);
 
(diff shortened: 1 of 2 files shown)


Coeur:
Oh, actually I just found a bug in **miniunz** for non-Windows! It will extract files with a timestamp of year 3917.

nmoinvaz:
Good find. I think I would consider this a bug and we should just fix it and not provide conversion function for the bug.  The minzip contrib that is part of zlib had its own time structure called tm_unz_s and that had defined year as starting at 1980. I think when I got rid of that structure, I didn't realize that difference between it and tm.

PR #144, opened 2017-09-19
allowing AES to be optional at runtime

Coeur:
When we build minizip with AES support, then [line 981 of zip.c](https://github.com/nmoinvaz/minizip/blob/master/zip.c#L981) is forcing AES encryption.

To give the choice at runtime, I made `zipOpenNewFileInZip5_64` with distinct parameters `method` and `compression_method`. This way, existing `zipOpenNewFileInZip4_64` behavior is unaffected.

### DIFF
--- a/minizip.c
+++ b/minizip.c
@@ -50,16 +50,17 @@ void minizip_banner()
 
 void minizip_help()
 {
-    printf("Usage : minizip [-o] [-a] [-0 to -9] [-p password] [-j] file.zip [files_to_add]\n\n" \
+    printf("Usage : minizip [-o] [-a] [-0 to -9] [-N] [-p password] [-j] file.zip [files_to_add]\n\n" \
            "  -o  Overwrite existing file.zip\n" \
            "  -a  Append to existing file.zip\n" \
            "  -0  Store only\n" \
            "  -1  Compress faster\n" \
-           "  -9  Compress better\n\n" \
+           "  -9  Compress better\n" \
+           "  -N  No AES\n\n" \
            "  -j  exclude path. store only the file name.\n\n");
 }
 
-int minizip_addfile(zipFile zf, const char *path, const char *filenameinzip, int level, const char *password)
+int minizip_addfile(zipFile zf, const char *path, const char *filenameinzip, int level, const char *password, int no_aes)
 {
     zip_fileinfo zi = { 0 };
     FILE *fin = NULL;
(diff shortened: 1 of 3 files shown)


Coeur:
@nmoinvaz I'm open to suggestions, to do it differently or to even rewrote our public headers more and call it "version 1.3.0" or something.

My notes:

* The CRC became useless, so those functions should be deprecated:
** zipOpenNewFileInZip3
** zipOpenNewFileInZip3_64
** zipOpenNewFileInZip4
** zipOpenNewFileInZip4_64

* We want to have AES optional at runtime:
** either by having `method` and adding a `compression_method` parameter (current pull request)
** either by having `method` and adding an `AES` true/false parameter

A third option would be to stop having so many function parameters and instead use a pointer to a curfile64_info.

What do you think could be a good approach, Nathan?

Coeur:
Oh, I was blind, actually we already have `zipFile file` that has its `file->ci` that can convey information. So I'm seeing a potential way to reduce the number of parameters here.

nmoinvaz:
I haven't had a chance to take a look at it yet.

nmoinvaz:
I would be in favor removing some of the older functions and just replacing them all with a simple function that uses a new `zip_fileinfo` defined like so:
```
typedef struct
{
    uint32_t    dos_date;               /* ms-dos date and time */
    uint16_t    internal_fa;            /* internal file attributes */
    uint32_t    external_fa;            /* external file attributes */
    void        *extrafield_local;      /* extra fields in local header*/
    uint16_t    size_extrafield_local;  /* size of additional extra fields in local header */
    void        *extrafield_global;     /* extra fields in global header */
    uint16_t    size_extrafield_global; /* size of extra fields in global header */
    char        *comment;               /* file comment */
    uint32_t    crc_for_crypting;       /* crc to use for traditional encryption */
    uint16_t    version_madeby;         /* version of zip made by */
    uint8_t     zip64;                  /* enable zip64 extensions */
    uint8_t     raw;                    /* no compression method */
    uint8_t     winzip_aes;             /* enable winzip aes encryption */
} zip_fileinfo;

It makes a lot more sense to go forward like this. Maybe we can add a compatibility layer.

Coeur:
Sounds good. So, if I get it correctly:

* **zip_fileinfo** would be the public structure for a file to write in a zip
* **curfile64_info** would be the internal structure for a file to write in a zip

And using `zipOpenNewFileInZip` would transform the public representation to the internal representation?

nmoinvaz:
Yes I would be in favor of all those changes.

Coeur:
for `version_madeby`, if it's in the structure it will conflict with the `VERSIONMADEBY` macro.

Coeur:
@nmoinvaz new zip_fileinfo done... but actually as three structures:
* zip_fileinfo for stuff like filename
* zip_compressioninfo for compression method
* zip_cryptinfo for crypto method

I've kept compatibility with existing functions, but I marked them as deprecated.

nmoinvaz:
I have started the process of merging your changes into the dev branch.

nmoinvaz:
This is now available in the dev branch via the mz_zip_crypt.aes variable. Thanks so much for your work!

PR #146, opened 2017-09-26
Add support for non-ASCII characters in paths on Windows

Timac:
On Windows, paths containing non ASCII characters were not supported. The reason is that fopen on Windows only supports ASCII characters in paths.

This change adds a define WINDOWS_SUPPORT_UNICODE_PATH which uses _wfopen instead of fopen64 on Windows. Tested on Windows 10.

### DIFF
--- a/ioapi.c
+++ b/ioapi.c
@@ -27,6 +27,10 @@
 #  define snprintf _snprintf
 #endif
 
+#if defined(WINDOWS_SUPPORT_UNICODE_PATH)
+#include <Windows.h>
+#endif
+
 voidpf call_zopen64(const zlib_filefunc64_32_def *pfilefunc, const void *filename, int mode)
 {
     if (pfilefunc->zfile_func64.zopen64_file != NULL)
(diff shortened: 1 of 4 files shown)


nmoinvaz:
Thanks for this! It looks good. However, I think it may be best to not use a WINDOWS_SUPPORT_UNICODE_PATH #define, but instead it should work with unicode by default.

nmoinvaz:
You should be using iowin32.c on Windows not ioapi.c. I will keep the changes to minishared but not to ioapi. Thanks.

Timac:
Note that on Windows ioapi.c is internally used. Here is a backtrace when you call zipOpen64() on Windows:
`MyApp.exe!fopen64_file_func(void * opaque, const void * filename, int mode) Line 149  C
MyApp.exe!call_zopen64(const zlib_filefunc64_32_def_s * pfilefunc, const void * filename, int mode) Line 37   C
MyApp.exe!zipOpen4(const void * path, int append, unsigned __int64 disk_size, const char * * globalcomment, zlib_filefunc64_32_def_s * pzlib_filefunc64_32_def) Line 635  C
MyApp.exe!zipOpen3(const char * path, int append, unsigned __int64 disk_size, const char * * globalcomment, zlib_filefunc_def_s * pzlib_filefunc32_def) Line 887  C
MyApp.exe!zipOpen64(const void * path, int append) Line 911   C`

Right now support for non-ASCII characters in paths is still broken. I will make a new pull request.

Timac:
Following pull request can be seen here:
https://github.com/nmoinvaz/minizip/pull/151

PR #147, opened 2017-09-26
Fix unused parameter warnings generated by -Wunused-parameter

Timac:
Fix all the unused parameter warnings generated by -Wunused-parameter on macOS with Xcode 9.

### DIFF
--- a/ioapi.c
+++ b/ioapi.c
@@ -81,7 +81,7 @@ void fill_zlib_filefunc64_32_def_from_filefunc32(zlib_filefunc64_32_def *p_filef
     p_filefunc64_32->ztell32_file = p_filefunc32->ztell_file;
 }
 
-static voidpf   ZCALLBACK fopen_file_func(voidpf opaque, const char *filename, int mode);
+static voidpf   ZCALLBACK fopen_file_func(ZIP_UNUSED voidpf opaque, const char *filename, int mode);
 static uint32_t ZCALLBACK fread_file_func(voidpf opaque, voidpf stream, void* buf, uint32_t size);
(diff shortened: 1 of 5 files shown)


Coeur:
are you on Windows or macOS? ah ah 👍

Timac:
Mostly macOS but Windows from time to time :)

nmoinvaz:
I was afraid of playing the cat and mouse game with unused parameters warning. I think it might be better to just disable the parameter warning instead.

Timac:
For my project, I want to keep the unused parameters warning enabled. Also for reasons I can't discuss here, I can't ignore -Wunused-parameter just for certain files.

So I have 2 solutions:
- I add ZIP_UNUSED to each unused parameter as in this pull request
- I tell clang to ignore the -Wunused-parameter warnings at the start of the file with
`#pragma clang diagnostic ignored "-Wunused-parameter"`

I personally find it cleaner to use ZIP_UNUSED. However the #pragma clang diagnostic solution would be easier to maintain if you really want to just disable the parameter warning.

Timac:
If you really want to disable the warnings, you could use something like:
`#if defined(__clang__)
	#pragma clang diagnostic ignored "-Wunused-parameter"
#elif _WIN32
	#pragma warning(disable : 4100)
#endif`

I can make such a pull request if you want.

Coeur:
I was the one to introduce `ZIP_UNUSED` with #126 and I would prefer to keep it. So I approve this pull request: it gives us a hint that **in a future version** we can refactor without the `opaque` parameter.

PR #148, opened 2017-09-26
Clear the tm struct properly to avoid a compiler warning

Timac:
Clear the tm struct properly to avoid compiler warning "missing field 'tm_min' initializer" when using -Wmissing-field-initializers


nmoinvaz:
There is probably no need for the memset, since we already use memcpy below.

Timac:
Thanks for the feedback. I made a simpler fix. See https://github.com/nmoinvaz/minizip/pull/150
I am closing this pull request.

Coeur:
@Timac, note that you can always rewrite a branch without creating a new pull request.

PR #151, opened 2017-09-27
Add support in ioapi.c for non-ASCII characters in paths on Windows

Timac:
On Windows, paths containing non ASCII characters are not supported. If you call the zipOpen64() function on Windows, it will internally call fopen64_file_func() in ioapi.c:

> MyApp.exe!fopen64_file_func(void * opaque, const void * filename, int mode) Line 149  C
> MyApp.exe!call_zopen64(const zlib_filefunc64_32_def_s * pfilefunc, const void * filename, int mode) Line 37   C
> MyApp.exe!zipOpen4(const void * path, int append, unsigned __int64 disk_size, const char * * globalcomment, zlib_filefunc64_32_def_s * pzlib_filefunc64_32_def) Line 635  C
> MyApp.exe!zipOpen3(const char * path, int append, unsigned __int64 disk_size, const char * * globalcomment, zlib_filefunc_def_s * pzlib_filefunc32_def) Line 887  C
> MyApp.exe!zipOpen64(const void * path, int append) Line 911   C

fopen64_file_func() uses fopen which does not support unicode paths on Windows. In order to support non ascii characters on Windows, fopen64_file_func() needs to be modified to not call fopen directly.

### DIFF
--- a/ioapi.c
+++ b/ioapi.c
@@ -27,10 +27,6 @@
 #  define snprintf _snprintf
 #endif
 
-#ifdef _MSC_VER
-#include <Windows.h>
-#endif
-
 voidpf call_zopen64(const zlib_filefunc64_32_def *pfilefunc, const void *filename, int mode)
 {
     if (pfilefunc->zfile_func64.zopen64_file != NULL)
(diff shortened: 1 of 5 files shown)


Timac:
This is a new pull request following https://github.com/nmoinvaz/minizip/pull/146

nmoinvaz:
Instead of using fill_fopen64_filefunc you should be using fill_win32_filefunc.

Timac:
I tried to use fill_win32_filefunc instead of fill_fopen64_filefunc. This works only if IOWIN32_USING_WINRT_API is defined and the UNICODE code not used. I have no idea if this is correct. Any help or info on that? Thanks!

nmoinvaz:
If you want Windows unicode support, you have to pass in a reference to a zlib_filefunc64_def that is filled by calling fill_win32_filefunc to either zipOpen2 and zipOpen2_64. I would rather not have references to iowin32.h in zip.c or unzip.c which for the most part are intended to be platform independent and only handle the zip file format functionality.

PR #156, opened 2017-10-01
compilation fix when using c99 or c11

Coeur:
also adding default compilation warning flags for the xcode project


Coeur:
@nmoinvaz  I saw you removed min/max in 1477a61ef81433f63f0979eb2d3366e9d5772584.
yet, this pull request is safe to merge to master to benefit from the xcodeproj changes (and to fix c99 before switching to streams).

PR #160, opened 2017-10-04
considering raw is always true to workaround incorrect headers

Coeur:
Fix https://github.com/nmoinvaz/minizip/issues/159


Coeur:
This pull request has consequences on large zip files. I do not know if it's good or bad, but a lower amount of data is actually extracted from #184 sample on this branch.

Coeur:
Confirmed, this change breaks something on the file from #184. :(

Coeur:
New PR with a better fix: #187

PR #164, opened 2017-10-05
Podspec for 1.3.0

Coeur:
I've removed the subspec for minishared, as apparently it's tools are now included in mz_strm.h.

### DIFF
--- a/Minizip.podspec
+++ b/Minizip.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name     = 'Minizip'
-  s.version  = '1.2.0'
+  s.version  = '1.3.0'
   s.license  = 'zlib'
   s.summary  = 'Minizip contrib in zlib with latest bug fixes that supports PKWARE disk spanning, AES encryption, and IO buffering'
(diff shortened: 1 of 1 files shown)


Coeur:
oups, wrong destination branch, and I don't know how to edit that

Coeur:
@nmoinvaz Note that from http://semver.org/ recommendations, by renaming the files and removing old functions signatures, we're doing breaking changes, so it should be numbered Minizip 2.0.0 instead...

Also, I would be in favor of adding a compatibility layer for old functions and mark them as deprecated before merging to Master branch.

nmoinvaz:
I had a compatibility layer but I removed it because I changed the way the open functions work but I suppose I could make a structure and encapsulate the extra stream needed. 
https://github.com/nmoinvaz/minizip/commit/4166027081fac15d8658d3675d8e88893c4070a8

nmoinvaz:
Also I agree with changing the version to 2.0.0

nmoinvaz:
I updated the version to 2.0.0 and will add a compatibility layer soon.

PR #172, opened 2017-10-06
guarding from missing path

Coeur:


Coeur:
Hum, actually my change is no good... I can't process command line options if I return. Let me rework this.

nmoinvaz:
Thanks!

PR #176, opened 2017-10-06
proper min/max

Coeur:


nmoinvaz:
Reference: https://stackoverflow.com/questions/3437404/min-and-max-in-c

nmoinvaz:
__typeof__ des not appear to be supported on MSVC.

Coeur:
using `decltype` instead of `__typeof__` when it's WINDOWS should solve it. But I don't have any Windows machine to verify. Could you try?

nmoinvaz:
It doesn't appear to be working, because decltype is a newer thing and I think only part of C++11.

Coeur:
OK, then let's drop this pull request, it's not that important.

PR #180, opened 2017-10-07
avoiding **file_info

Coeur:
Not sure why we would need a double pointer for file_info.

### DIFF
--- a/miniunz.c
+++ b/miniunz.c
@@ -51,7 +51,7 @@ void miniunz_help()
 
 int32_t miniunz_list(void *handle)
 {
-    mz_unzip_file *file_info = NULL;
+    mz_unzip_file file_info;
     uint32_t ratio = 0;
     int16_t level = 0;
     int16_t err = MZ_OK;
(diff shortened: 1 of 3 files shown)


nmoinvaz:
I've never really used structs this way and don't understand why it is working - I would think that a memcpy would be necessary inside get_file_info, but it seems to be fine -- ie. `memcpy(file_info, &unzip->file_info, sizeof(mz_unzip_file_info))`. I would usually just return a pointer to the already existing structure.

Coeur:
By writing `mz_unzip_file file_info;`, space for the struct is auto allocated by the caller function (for example `miniunz_list()`). In C, a copy is always done by `*file_info = unzip->file_info;`, no need for memcpy. (https://stackoverflow.com/a/9127315/1033581)

By writing `mz_unzip_file *file_info = NULL;`, space for the struct is allocated by the called function (`mz_unzip_entry_get_info()`). No copy is done by `*file_info = &unzip->file_info;`, so if the struct is modified, it will have potentially side effects...

nmoinvaz:
That makes sense. Thanks for the StackOverflow article. I guess I haven't really seen that usage that much all these years. You mentioned pontential side effects of the struct being modified... the memory pointed to by the `filename`, `extrafield` and `comment` parameters of the structure will change when the user moves to the next file using to goto_first/goto_next functions. It might be better to keep it as a double pointer so  that people know that it is only valid for the current file. I can add a comment in to the same effect. What do you think?

Coeur:
Well, I don't know which one is the correct solution. They have their pros and cons.

Sounds to me that:
* using *file_info is safer
* using **file_info may be slightly optimized

Is it possible to do `*const*file_info`?

nmoinvaz:
I think I am decided to leave it as is. Changing it to the way you suggested might cause people to think that all the file info data is being copied - even though the data pointed to by `filename`, `extrafield` and `comment` are not copied. I would rather not do a half copy of the associated file info data. I have added a comment into the mz_unzip.h to that affect to make it clear that the file info is only valid while the current file in the zip is navigated to.

PR #186, opened 2017-10-11
Compatibility with archives of more than 65535 files without Zip64 support

### LINKED ISSUE #184: can't fully extract a zip file with more than 65535 files
Problem reported by @jdeepwell.
http://deepwell.name/Apple_API_Reference.docset.zip is a 968.6 MB archive created by a native macOS tool.

Normally, once uncompressed, it should be roughly 2,129,626,642 bytes (~ 2,031 MB).
But miniunz (from 1.2 branch) will only extract maybe ~298.6 MB and will not report any error. For instance, file `hc8z3wx_vy.html` will be missing when unzipping with miniunz.

To me it seems that `unzReadCurrentFile()` may return Z_BUF_ERROR at some point.

Coeur:
Fix #184

### DIFF
--- a/unzip.c
+++ b/unzip.c
@@ -350,6 +350,7 @@ static unzFile unzOpenInternal(const void *path, zlib_filefunc64_32_def *pzlib_f
     uint64_t value64 = 0;
     voidpf filestream = NULL;
     int err = UNZ_OK;
+    int err64 = UNZ_OK;
 
     us.filestream = NULL;
     us.filestream_with_CD = NULL;
@@ -412,7 +413,7 @@ static unzFile unzOpenInternal(const void *path, zlib_filefunc64_32_def *pzlib_f
         if (err == UNZ_OK)
         {
             /* Search for Zip64 end of central directory header */
-            int err64 = unzSearchCentralDir64(&us.z_filefunc, &central_pos64, us.filestream, central_pos);
+            err64 = unzSearchCentralDir64(&us.z_filefunc, &central_pos64, us.filestream, central_pos);
             if (err64 == UNZ_OK)
             {
                 central_pos = central_pos64;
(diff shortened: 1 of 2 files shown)


nmoinvaz:
I think that it might be better to not rely on number of entries in get first/next file functions. But it should reach the end when a read function returns a different magic value.

nmoinvaz:
When it reaches the magic value for one of these:

[central directory header 1]
      .
      .
      .
      [central directory header n]
      [zip64 end of central directory record]
      [zip64 end of central directory locator] 
      [end of central directory record]

nmoinvaz:
Or reaches the end of the file.

Coeur:
I wouldn't know how to implement what you are describing.
But recounting the number of entries allows for correct progress reporting (i.e.: 1/10000 unzipped, 2/10000 unzipped, 3/10000 unzipped, etc.)

nmoinvaz:
I see what you mean that the number count may be still off for people who are using that for progress. The zip file is not using the Zip64 extension which supports more than 64K files. The tool that was used to create it probably doesn’t support it either. 

The problem I have with the pull request is it does a lot of extra seeking and reading of the file for all zip files (a majority of which will be correct) for the one or two that are incorrect.

Coeur:
As the tool that doesn't support Zip64 for compression is the native compression tool of macOS (so used by everybody by a simple right-click -> compress on any folder), we do need a workaround. And all known tools I have tried are compatible for unzipping correctly those files with wrong number_entry.

Now, all it does is seeking and verifying the magic number for each file, which is:
* less reading than listing all the files
* much less reading than unzipping all the files
* only being done on files without Zip64 support

So the overhead is low.

If we want to lower the overhead even more, we'll have to give an option to the user of minizip to skip recounting the files if he wishes so. If we go that way and make the recounting optional, may I know what would be the signature of the method? Some `unzOpenCurrentFile4`?

nmoinvaz:
This would only fix issues with zip files that are less than 4GB that have more than 64K files. Using the native zipper on mac would still create an invalid archive if it included files that were greater than 4GB or if the zip file itself was greater than 4GB some of the offsets might be wrong. For this issue, I am only willing to work around the fact that we use number_entry in the global info structure. The unzipper would be able to extract all 64K+ files in this instance, but the number_entry would still be incorrect. I will leave it up to the person integrating the library to do the additional check to see if number_entry is correct, if that is what they want to do.

PR #187, opened 2017-10-11
workaround incorrect headers (fix #159)

### LINKED ISSUE #159: Support correct unzipping of IncorrectHeaders.zip
File with Incorrect Headers:
[IncorrectHeaders.zip](https://github.com/nmoinvaz/minizip/files/1356685/IncorrectHeaders.zip)

Unzipping it is supposed to give a 59 bytes file ending by _Julius Parishy_.

Applications that succeed at unzipping it on macOS:
* /usr/bin/unzip 6.00 of 20 April 2009 succeeds
* Archive Utility 10.10 (81) succeeds
* The Unarchiver 3.11.3 (115) succeeds

Applications that fail at unzipping it on macOS:
* MiniUnz 1.2.0 fails, as resulting file is truncated to 50 bytes and is ending by _Julius_ only. Logs:
>$ ./miniunz IncorrectHeaders.zip 
MiniUnz 1.2.0, demo of zLib + Unz package
more info at https://github.com/nmoinvaz/minizip
>
>IncorrectHeaders.zip opened
creating directory: IncorrectHeaders/
 extracting: IncorrectHeaders/Readme.txt
error -5 with zipfile in unzReadCurrentFile
error -105 with zipfile in unzCloseCurrentFile
error -5 with zipfile in unzGoToNextFile

Coeur:
Proper fix for #159.

### DIFF
--- a/unzip.c
+++ b/unzip.c
@@ -1297,16 +1297,11 @@ extern int ZEXPORT unzReadCurrentFile(unzFile file, voidp buf, uint32_t len)
     s->pfile_in_zip_read->stream.next_out = (uint8_t*)buf;
     s->pfile_in_zip_read->stream.avail_out = (uint16_t)len;
 
-    if (s->pfile_in_zip_read->raw)
+    if ((s->pfile_in_zip_read->compression_method == 0) || (s->pfile_in_zip_read->raw))
     {
         if (len > s->pfile_in_zip_read->rest_read_compressed + s->pfile_in_zip_read->stream.avail_in)
             s->pfile_in_zip_read->stream.avail_out = (uint16_t)s->pfile_in_zip_read->rest_read_compressed +
-                s->pfile_in_zip_read->stream.avail_in;
-    }
-    else
-    {
-        if (len > s->pfile_in_zip_read->rest_read_uncompressed)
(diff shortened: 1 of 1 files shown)


nmoinvaz:
It looks like that if statement is in the version of minizip that is distributed with zlib. Also, if compressed size is wrong in the header it will also have problems?

Coeur:
As the issue says, those three tools are compatible with the _incorrect headers_ files:
* /usr/bin/unzip 6.00 (native on macOS)
* Archive Utility 10.10 (native on macOS)
* The Unarchiver 3.11.3

The analysis of the original fix made by Julius Parishy (@jparishy) was:
>maybe Minizip is using the cached uncompressed sizes in the file (which in both the local header and central directory is noted as 1857, when in fact it should be 1860) while other tools are uncompressing the data manually and getting the size directly from the outputted data.

@nmoinvaz  If you give me a sample with incorrect compressed size, I can test the behavior on all the tools I know.

nmoinvaz:
I looked into it and an incorrect compressed size will not work with AES because AES needs to know the amount of bytes to crunch for calculating the HMAC-SHA1.

PR #190, opened 2017-10-17
Fix compilation when NOCRYPT or NOUNCRYPT is defined

### LINKED ISSUE #189: Compilation fails if NOCRYPT or NOUNCRYPT is defined
When NOCRYPT is defined, the conditional code inside mz_zip_entry_open() does not compile.  The compiler complains that the "password" variable is not defined.

When NOUNCRYPT is defined, the conditional code inside mz_unzip_entry_open() fails to compile due to mismatched brackets.

timothytylee:
This fixes #189


nmoinvaz:
Thank you, this should be fixed now.

PR #193, opened 2017-10-17
"mz_error.h" no longer exists

timothytylee:
mz_compat.c fails to compile because "mz_error.h" no longer exists.  It should include "mz.h" instead.


nmoinvaz:
Thanks again!

PR #194, opened 2017-10-17
Include "errno.h" to define EEXIST

timothytylee:
When compiling under mingw-w64, "errno.h" must be included to define EEXIST


nmoinvaz:
Thanks!

PR #206, opened 2017-11-11
Fixes for the mz_compat layer.

GerbilSoft:
While trying to update my ROM Properties Page plugin to use MiniZip 2.2.3, I encountered several issues in the mz_compat layer that caused crashes and other random problems. One of these problems were already fixed here before I had the chance to send the PR (build without encryption).

This PR has:
* Correctly dereference `pzlib_filefunc_def`. This is only an issue if the MiniZip 1.x vtable is set manually, which I was doing in order to handle Unicode filenames on Windows.
* unzClose(): Close the file handle. A similar fix for zipClose2_64() was applied in commit 053d00c451e102464e680069cd1ed6cae4171dc6.
* unzGetCurrentFileInfo64(): Make sure `filename` and `comment` are NULL-terminated.
* unzOpen2_64(): Go to the first file immediately.
* Some printf() fixes found with cppcheck-1.81.

### DIFF
--- a/src/minizip.c
+++ b/src/minizip.c
@@ -595,7 +595,7 @@ int main(int argc, char *argv[])
                 if ((c == 's') || (c == 'S'))
                     options.aes = 1;
 #endif
-                if (((c == 'k') || (c == 'k')) && (i + 1 < argc))
+                if (((c == 'k') || (c == 'K')) && (i + 1 < argc))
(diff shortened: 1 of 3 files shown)


nmoinvaz:
Great fixes! Thanks for your contribution!

PR #207, opened 2017-11-12
mz_compat.c: Fix zipOpen(), zipOpen64(), and unzOpen64().

GerbilSoft:
The previous bugfixes fixed the functions that took a zlib_filefunc_def* parameter, but they inadvertently broke the functions that didn't. The fix is to store the filefunc pointer locally, then pass a pointer to the pointer.

### DIFF
--- a/src/mz_compat.c
+++ b/src/mz_compat.c
@@ -34,12 +34,14 @@ typedef struct mz_compat_s {
 
 extern zipFile ZEXPORT zipOpen(const char *path, int append)
 {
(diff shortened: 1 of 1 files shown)


GerbilSoft:
Also, on Debian/Ubuntu, `libbsd-dev` is needed in order to use <bsd/stdlib.h>. libbsd has a pkgconfig file, so FIND_PACKAGE(libbsd REQUIRED) should work here.

nmoinvaz:
Thanks! Do you want to submit another pull request for the libbsd change?

PR #209, opened 2017-11-18
Mark libbsd as REQUIRED on Linux.

GerbilSoft:
MiniZip won't compile properly if libbsd is missing, but the current pkg_check_modules() call doesn't enforce this. This PR adds the REQUIRED keyword if compiling for Linux. It also skips the libbsd check on non-Linux systems.

mz_os_posix.c: Implement arc4random_buf() for Mac OS X earlier than 10.7. 10.6 has arc4random(), but not arc4random_buf(). (Note that this code is untested.)


Coeur:
Forget about testing `< MAC_OS_X_VERSION_10_7`...
* Mac OS X 10.7 Lion was released on July 20, 2011 and latest security update for 10.6 Snow Leopard was on July 25, 2011, so Mac OS 10.6 became obsolete more than 6 years ago.
* Mac OS developers generally use weak properties, which requires Mac OS 10.7 anyway.
* Mac OS developers generally use Swift dependencies, which requires Mac OS 10.9 anyway.
* Mac OS developers generally use Storyboards, which requires Mac OS 10.10 anyway.

nmoinvaz:
Thanks for the pull request. I am currently not able to take a look at it for some time, but I will take a look at it as soon as I get a chance.

GerbilSoft:
The only Mac I have runs 10.5 (Power Mac G4 MDD), which is why I included the check. I can remove it if it doesn't make sense to include it anymore.

nmoinvaz:
I agree with @Coeur, it probably doesn't make sense to add a shim for arc4random_buf. If you remove it I can merge the pull request.

nmoinvaz:
Thanks for your work!

PR #210, opened 2017-12-07
Verify decrypted header to detect wrong password for Traditional PKZIP.

yamaguchi-am:
This change adds a quick check for the password upon file open, for Traditional PKZIP encryption.
It will avoid that the applications continue to read incorrect data when wrong password is entered.
The verification logic is noted in the documents published by http://www.info-zip.org/doc/.

We would like to have this fix in the 1.2 branch because Zip Archiver extension for ChromiumOS depends on it.
https://bugs.chromium.org/p/chromium/issues/detail?id=782197


nmoinvaz:
Thanks! I have also made the appropriate changes for the master branch.

PR #211, opened 2017-12-18
Allow building without libbsd on Linux when NOCRYPT symbol is defined.

yamaguchi-am:


yamaguchi-am:
This patch will stop compiling 2 unused functions in case of NOCRYPT symbol defined.
IIUC the arc4random_buf being used by those is the only place it requires libbsd on Linux systems.
We would like to have this fix in the 1.2 branch because Zip Archiver extension for ChromiumOS (which doesn't support encryption, but just decryption) depends on it, and we'd like to minimize the dependency libs.

PR #214, opened 2018-01-01
mz_strm_posix: Fix return value checking for POSIX stdio functions

chenxiaolong:
This PR fixes a few functions in `mz_strm_posix.c` to check for the right failure conditions in the POSIX stdio API.


nmoinvaz:
Thank you!

PR #215, opened 2018-01-01
CMakeLists.txt: Use add_compile_options() instead of modifying CMAKE_C_FLAGS

chenxiaolong:
This allows a parent project to use minizip via:

    add_subdirectory(path/to/minizip)

without the compiler flags of the parent project being affected.


nmoinvaz:
Thanks!

PR #219, opened 2018-01-04
Cmake fixes

mikecrowe:
These fixes should make it easier to install minizip as a system library on Linux. I'm not sure what effect they will have on Windows or MacOS, but I would hope that GNUInstallDirs does something sensible there.


nmoinvaz:
Looks good thanks for the contribution.

PR #220, opened 2018-01-06
mz_zip.c: Fix incorrect size of data descriptor fields

chenxiaolong:
Previously, the conditions for writing 32-bit vs 64-bit data descriptor
fields were swapped. If the size was smaller than UINT32_MAX, a 64-bit
integer would be written, and if the size was larger, then a truncated
32-bit integer would be written. This prevented programs that don't have
support for the central directory (like busybox) from reading files
created by minizip.


chenxiaolong:
This also appears to fix an issue where 7zip cannot edit zip files created by minizip. (Previously, it would fail with `System ERROR: E_NOTIMPL`)

nmoinvaz:
Great find! Thank you.

PR #222, opened 2018-01-07
cmake: Use POSITION_INDEPENDENT_CODE property instead of setting -fPIC

chenxiaolong:
CMake will figure out the right compiler arguments to use for enabling
PIC on the target platform.


nmoinvaz:
Thanks!

PR #223, opened 2018-01-07
cmake: Enable x86 optimizations only if supported

chenxiaolong:
Previously, these optimizations would get enabled when building for
non-x86 architectures using Clang, which would cause a warning to be
output.


nmoinvaz:
Thanks!

PR #224, opened 2018-01-09
mz_zip.c: Both data descriptor size fields must be the same type

chenxiaolong:
Section 4.3.9.1 of the standard says:

    For ZIP64(tm) format archives, the compressed and uncompressed
    sizes are 8 bytes each.

This updates the data descriptor code to use 64 bit integers if either
the compressed size or the compressed size is greater than UINT32_MAX.


chenxiaolong:
Closing this PR. Sorry, I misread the original code, which was already correct.

PR #231, opened 2018-02-05
keep FILE_SHARE_READ if MZ_OPEN_MODE_READ

rip-nsk:
mz_stream_win32_open opens the stream in exclusively mode if mode & MZ_OPEN_MODE_READWRITE) == MZ_OPEN_MODE_READ.
Likely it is a bug, and it is not clear what is expected behavior.
Just keep FILE_SHARE_READ to allow shared read access to the stream.


nmoinvaz:
Thanks!

PR #234, opened 2018-02-08
ZIP64 format is required if number of entries in archive is greater than 65535.

juanii:
Previous to this change, if there was more than 65535 files in an archive, the CD fields "total number of entries on this disk" and "total number of entries in CD" were properly set to 0xFFFF but no ZIP64 CD was written with the proper number of entries in archive.


juanii:
Commit 30589b5 addresses a problem with the "version made by" field set by the minizip sample program and the compat module. The field was being set just to the "host system" code (which is dependent of the compilation platform).
In my case (compiling on macOS, code 19) the zipinfo utility listed files inside an archive as being created by a program supporting (nonexistent) ZIP version 1.9 in an MS-DOS host system (code 0, the empty lower byte).
This commit moves the "host system" code to the upper byte of the "version made by" field and sets the ZIP version (lower byte) depending on the compilation flags: 4.5 by default, 4.6 if BZIP2 is enabled, 5.1 if AES is enabled and 6.3 if LZMA is enabled.

juanii:
Commit c65486d fixes the value for the "total number of disks" in ZIP64 end of central directory locator. Currently the value is set to the contents of the disk_number_with_cd variable. This confuses ZipInfo/Unzip utilities to the point of being unable to properly calculate the central directory offset, making it impossible to process these archives in any way.

Since minizip does not seem to allow central directory splitting*, this quick fix sets the "total number of disks" to the value of disk_number_with_cd (which seems to be zero-based) plus one, yielding the correct number of split volumes. In the case of single-volume archives, since split streams are not used and OS/buffered streams do not respond to the query of MZ_STREAM_PROP_DISK_NUMBER property, the disk_number_with_cd is always 0 and total number of disks is 1.

There's a special case in the source code for split volumes with disk_size equal to zero which I don't understand very well and modifies the value in disk_number_with_cd. I tested it with minizip (-k 0) and this fix worked fine.

* At least in my tests, please comment if I'm wrong.

nmoinvaz:
The fix for the bug about the number of entries look right. The version made by solution is a good one. For the number of disks issue, the older 1.2 branch also had a similar logic, https://github.com/nmoinvaz/minizip/blob/53a657318af1fccc4bac7ed230729302b2391d1d/zip.c#L1958 so I am comfortable with the change. Thank you!

juanii:
You're welcome. I really appreciate your work maintaining such an useful and fundamental piece of software. I'm looking forward to make more contributions. Whenever I have some free time I'd like to implement UTF-8 file names and comments. Thank you.

PR #236, opened 2018-02-10
Fix: unzGetGlobalInfo failed with MZ_EXIST_ERROR if there was no comment in ZIP file.

juanii:


nmoinvaz:
Thanks!

juanii:
You're welcome. But watch out! It's failing again after commit c5e2d27. The affected code is:

https://github.com/nmoinvaz/minizip/blob/c0cc7df7432f95fdc586ad2b1ab0c29ee7970dc7/src/mz_compat.c#L348-L355

The call to `mz_zip_get_comment` on an archive without a global comment returns `MZ_EXIST_ERROR`. The size_comment assignment is correctly skipped, but the following code to set the other struct members is not executed, and an error is returned to the user. I think the `MZ_EXIST_ERROR` should only cause size_comment to be zero and then be silently discarded. That's how the 1.x version worked.

Regards

nmoinvaz:
Sorry I didn't see that. Do u want to make a new pull request? Thanks.

PR #238, opened 2018-02-11
Fix ntfs extra field

juanii:
Currently the implementation for NTFS file times extra field writes only the timestamps different from zero:

https://github.com/nmoinvaz/minizip/blob/7a3820c53912d3c004c493b0ad1811864ccfe8c1/src/mz_zip.c#L883-L889

From the APPNOTE.TXT I understand that the extra field should contain all the three timestamps (or not be used at all). Since I'm no expert, I peeked at a few ZIP libraries/utilities source code. Multiplatform utilities (like InfoZIP) usually skip the field altogether or do not parse its contents. So I focused on libraries targeting the Windows system and (IMHO) are the most relevant. To cite some examples, SharpZipLib will ignore the extra field if tag data length is different from 24 bytes. DotNetZip will fail with an exception if the extra field total length is different from 32 bytes. Other less popular libraries will do the same, either ignore the field or fail if the field is not complete with the three timestamps.

This change will write an NTFS extra field only if `modified_date`, `accessed_date` and `creation_date` are different from zero. Similarly, when reading an NTFS extra field, it will only be parsed if the three timestamps are present, otherwise the field is ignored, so older archives which may contain less than three timestamps won't break when trying to decompress.

Finally, `zipOpenNewFileInZip5` in `mz_compat.c` now initializes `file_info` struct to zero. Before this change, since the `if` body cited below only initializes `modified_date`, the variables `accessed_date` and `creation_date` have a great chance of containing non-zeroed values which will be written to the extra field, potentially breaking other decompressors which parse and try to convert those values to timestamps.

https://github.com/nmoinvaz/minizip/blob/7a3820c53912d3c004c493b0ad1811864ccfe8c1/src/mz_compat.c#L116-L121

Regards

### DIFF
--- a/src/mz_compat.c
+++ b/src/mz_compat.c
@@ -106,7 +106,7 @@ extern int ZEXPORT zipOpenNewFileInZip5(zipFile file, const char *filename, cons
     ZIP_UNUSED uint32_t crc_for_crypting,  uint16_t version_madeby, uint16_t flag_base, int zip64)
 {
     mz_compat *compat = (mz_compat *)file;
-    mz_zip_file file_info;
+    mz_zip_file file_info = { 0 };
 
     if (compat == NULL)
         return MZ_PARAM_ERROR;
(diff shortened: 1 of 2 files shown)


nmoinvaz:
Thanks for the pull request.

PR #240, opened 2018-02-12
Fixes #233

### LINKED ISSUE #233: lzma data maybe don't have LZMA_STREAM_END flag.
you may use file_info->uncompressed_size to limit avail_out.

 read = mz_zip_entry_read(handle, buf, sizeof(buf));
->int32_t mz_stream_lzma_read(void *stream, void *buf, int32_t size)

->read = mz_stream_read(lzma->stream.base, lzma->buffer, bytes_to_read);
-> err = lzma_code(&lzma->lstream, LZMA_RUN);
err is 9 or 0xb

juanii:
Fixes decompressing LZMA streams without EOS marker, which is indicated by bit 1 turned off in general purpose bit flag (APPNOTE.TXT section 4.4.4).
I have absolutely no experience using LZMA library so it would be good if someone could check this. The only documentation I found was the Javadoc for the XZ library, which states that when EOS is missing, the library user must know both the compressed size to limit the compressed stream reads, and the uncompressed size to limit the output buffer writes, otherwise the decompressor might produce more output than the original (e.g., when decompressing nices/DirectX-setup/DSETUP.dll from the test file, an extra 0x00 byte is produced).

### DIFF
--- a/src/mz_strm_lzma.c
+++ b/src/mz_strm_lzma.c
@@ -55,6 +55,7 @@ typedef struct mz_stream_lzma_s {
     int64_t     total_in;
     int64_t     total_out;
     int64_t     max_total_in;
+    int64_t     max_total_out;
     int8_t      initialized;
     uint32_t    preset;
 } mz_stream_lzma;
@@ -181,6 +182,8 @@ int32_t mz_stream_lzma_read(void *stream, void *buf, int32_t size)
 
         total_in_after = lzma->lstream.avail_in;
         total_out_after = lzma->lstream.total_out;
+        if (lzma->max_total_out != -1 && total_out_after > lzma->max_total_out)
+            total_out_after = lzma->max_total_out;
         
         in_bytes = (uint32_t)(total_in_before - total_in_after);
         out_bytes = (uint32_t)(total_out_after - total_out_before);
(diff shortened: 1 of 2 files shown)


nmoinvaz:
Wouldn't it be..
``
 if ((lzma->max_total_out != -1) && (lzma->total_out + total_out_after > lzma->max_total_out))
            total_out_after = lzma->max_total_out - lzma->total_out;
``

juanii:
I can't be authoritative about this. As mentioned, I never used liblzma before but I think the `total_out_after` variable (which is directly assigned from `lzma_stream.total_out` in line 184) accumulates the total output across (potentially multiple) `lzma_code()` runs[1]. The current run's output is calculated as `total_out_after - total_out_before` in line 189, which is in turn added to `total_out` in line 192. But I might be missing something here (why `lzma->total_out` is not directly assigned with `lzma->lstream.total_out`? maybe it's just for readability, but it could be something I'm not seeing).

One thing I'm sure: the code worked for the sample file attached to the issue. I decompressed it with 7zip and the patched version of minizip and both results were identical. I know, it's just _one_ test case.

[1] I saw this while debugging the decompression of DSETUP.dll from the sample file, which requires two runs of `lzma_code()`. The input (compressed) stream fits entirely in `lzma->lstream.next_in` but produces uncompressed output that overflows `lzma->lstream.next_out`, and a second run is needed. And across these runs, `total_out_after` always increases.

nmoinvaz:
I think I am going to have to wait until I have more time and feeling better to investigate this.

nmoinvaz:
Looks good. Thank you!

PR #242, opened 2018-02-14
Fix compat layer's unzGetGlobalInfo() for archives without global com…

juanii:
…ment.


nmoinvaz:
I'm not sure the else if is necessary since the memset will reset the value to 0.

PR #248, opened 2018-03-18
Fix comment length overflow.

cosmy1:
In MiniZip compatible layer there is an overflow.
```cpp
mz_zip_get_disk_number_with_cd(compat->handle, (int64_t *)&pglobal_info->number_disk_with_CD);
```
unz_global_info64_s struct is declared as
```cpp
uint64_t number_entry; 
uint32_t number_disk_with_CD;
uint16_t size_comment;
```
number_disk_with_CD must be 64-bit length otherwise will overflow into the other struct member.


nmoinvaz:
I think `number_disk_with_CD` should remain uint32 for compatibility. I have fixed it another way. Let me know if you still have issues.

PR #251, opened 2018-04-22
improve compatibility with original minizip

vszakats:
* Fix NULL pointer dereference when calling `zipOpen*()` or `unzOpen2*()` with NULL filefunc table.
* Rename `dos_date` to `dosDate` in zip/unzip structures to be compatible with original minizip.
* Add tm struct timestamps to zip/unzip structures and populate them with the `dosDate` value for unzip and use the value passed into it on zip, if `dosDate` is zero (mimicing original minizip behaviour.)
* Change `zipWriteInFileInZip()` return value to be a status code (`ZIP_OK` or `ZIP_ERRNO`) instead of the number of bytes written. This way it's minizip compatible.
* Add `zipCloseFileInZipRaw64()` compatibility wrapper.
* Add these compatibility functions:
  `unzGetOffset()`
  `unzGetOffset64()`
  `unzSetOffset()`
  `unzSetOffset64()`
* Add stub for compatibility: `unzGetLocalExtrafield()`
   It will always return zero for now.
* Add original minizip types for compatiblity: `tm_unz`, `tm_zip`, `ZPOS64_T`
* Add these minizip (non-compatiblity) APIs:
  `mz_zip_get_entry()`
  `mz_zip_goto_entry()`
* `zipOpenNewFileInZip*()` to use the default filename of "-" if filename parameter was NULL. For minizip compatibility.
* Fix `mz_zip_entry_open_int()` and `mz_zip_entry_read_open()` failing with `MZ_PARAM_ERROR` in raw mode
* Add/cleanup some date-related comments
* `mz_zip_time_t_to_tm()` to clear the `tm` structure in case `localtime()` call failed.
* Fix `mz_zip_tm_to_dosdate()` to do the date validation on the normalized date (the year in particular), not on the raw value passed by the caller. This makes it work correctly when year is passed as-is, such as 1990.

The reference minizip was the version that comes with the latest zlib, currently at 1.2.11.

### DIFF
--- a/mz_compat.c
+++ b/mz_compat.c
@@ -59,8 +59,16 @@ extern zipFile ZEXPORT zipOpen2_64(const void *path, int append, const char **gl
     void *handle = NULL;
     void *stream = NULL;
 
-    if (mz_stream_create(&stream, (mz_stream_vtbl *)*pzlib_filefunc_def) == NULL)
-        return NULL;
+    if (pzlib_filefunc_def)
+    {
+        if (mz_stream_create(&stream, (mz_stream_vtbl *)*pzlib_filefunc_def) == NULL)
+            return NULL;
+    }
+    else
+    {
+        if (mz_stream_os_create(&stream) == NULL)
+            return NULL;
+    }
 
     switch (append)
     {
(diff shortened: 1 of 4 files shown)


vszakats:
These headers weren't included in the patch, but they are useful for full compatibility with traditional minizip 1.1:

`unzip.h`:
```c
#include "zip.h"
```
`zip.h`:
```c
#include "mz.h"
#include "mz_zip.h"
#include "mz_compat.h"
```

nmoinvaz:
Can you create the pull request against the dev branch? Thanks.

vszakats:
Done that. Closing this one in favour of https://github.com/nmoinvaz/minizip/pull/252.

PR #256, opened 2018-04-30
fix to set MZ_ZIP_FLAG_LZMA_EOS_MARKER flag in raw mode

vszakats:
in function `mz_zip_entry_write_open()`

Fixes https://github.com/nmoinvaz/minizip/issues/255


nmoinvaz:
Thank you!

PR #280, opened 2018-07-13
add missing EEXIST constant for mingw-w64

vszakats:


nmoinvaz:
Thanks! I have decided to include errno.h again.

vszakats:
`errno.h` is notably missing from Windows CE, so we're fine unless supporting this platform is a goal.

PR #282, opened 2018-07-16
restore DEF_MEM_LEVEL constant in mz_compat.h

vszakats:
This is defined in zutil.h in zlib, which is a private
header, so it doesn't get defined when #including zlib.h.
Instead if does get defined by minizip1 in zip.h, so
let's keep doing the same here as well.

Ref: https://github.com/madler/zlib/blob/v1.2.11/contrib/minizip/zip.h#L79-L85
Regression since: https://github.com/nmoinvaz/minizip/commit/ea99a46b6d36a11bc3e3eb3a806482d68ab4a991

### DIFF
--- a/mz_compat.h
+++ b/mz_compat.h
@@ -22,6 +22,13 @@ extern "C" {
 
 #ifdef HAVE_ZLIB
 #include "zlib.h"
+#ifndef DEF_MEM_LEVEL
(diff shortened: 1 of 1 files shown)


nmoinvaz:
Looks good. Thanks!

PR #283, opened 2018-07-17
add support for creating zip files in memory via the compatibility API

vszakats:
This adds 4 new, MZ-specific API calls to the compatibility interface:
```
void* zipGetStream_MZ(zipFile file)
int   zipClosePrepare_MZ(zipFile file, const char *global_comment)
int   zipClosePrepare2_MZ(zipFile file, const char *global_comment, uint16_t version_madeby)
int   zipCloseFree_MZ(zipFile file)
```

It also adds one API call to the native mz interface, making it
possible to query the type of a stream:
```
const void* mz_stream_get_interface(const void *stream)
```

### DIFF
--- a/mz_compat.c
+++ b/mz_compat.c
@@ -62,6 +62,10 @@ extern zipFile ZEXPORT zipOpen2_64(const void *path, int append, const char **gl
     {
         if (mz_stream_create(&stream, (mz_stream_vtbl *)*pzlib_filefunc_def) == NULL)
             return NULL;
+
+        // Configure a reasonable grow size if the stream is a memory one
+        if (pzlib_filefunc_def == mz_stream_mem_get_interface())
+            mz_stream_mem_set_grow_size(stream, 128 * 1024);
     }
     else
     {
@@ -280,6 +284,53 @@ extern int ZEXPORT zipClose2_64(zipFile file, const char *global_comment, uint16
     MZ_FREE(compat);
 
     return err;
+}
+
+extern void* ZEXPORT zipGetStream_MZ(zipFile file)
+{
+    mz_compat *compat = (mz_compat *)file;
+
+    return (void *)compat->stream;
+}
+
+// Similar to classic zipClose(), but will not destroy internal MZ
+// objects, e.g. allowing to retrieve the memory buffer via zipGetStream_MZ().
+// You must subsequently call zipCloseFree_MZ() to free those objects.
+extern int ZEXPORT zipClosePrepare_MZ(zipFile file, const char *global_comment)
+{
+    return zipClosePrepare2_MZ(file, global_comment, MZ_VERSION_MADEBY);
+}
+
+extern int ZEXPORT zipClosePrepare2_MZ(zipFile file, const char *global_comment, uint16_t version_madeby)
+{
+    mz_compat *compat = (mz_compat *)file;
+    int32_t err = MZ_OK;
+
+    if (compat == NULL)
+        return ZIP_PARAMERROR;
+
+    if (global_comment != NULL)
+        mz_zip_set_comment(compat->handle, global_comment);
+
+    mz_zip_set_version_madeby(compat->handle, version_madeby);
+    err = mz_zip_close(compat->handle);
+
+    return err;
+}
+
+extern int ZEXPORT zipCloseFree_MZ(zipFile file)
+{
+    mz_compat *compat = (mz_compat *)file;
+
+    if (compat->stream != NULL)
+    {
+        mz_stream_close(compat->stream);
+        mz_stream_delete(&compat->stream);
+    }
+
+    MZ_FREE(compat);
+
+    return MZ_OK;
 }
 
 /***************************************************************************/
--- a/mz_compat.h
+++ b/mz_compat.h
@@ -154,6 +154,11 @@ extern int ZEXPORT zipClose(zipFile file, const char *global_comment);
 extern int ZEXPORT zipClose_64(zipFile file, const char *global_comment);
 extern int ZEXPORT zipClose2_64(zipFile file, const char *global_comment, uint16_t version_madeby);
 
+extern void* ZEXPORT zipGetStream_MZ(zipFile file);
+extern int   ZEXPORT zipClosePrepare_MZ(zipFile file, const char *global_comment);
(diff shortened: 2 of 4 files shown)


vszakats:
Another way of doing this would be just add a single, MZ-specific compatibility API function to expose internal `handle` and `stream` pointers of the `compat` handle. This would allow to implement most of this logic in client code, without further touching minizip codebase and littering the compatibility interface with MZ-specific extensions. With these two pointers available one could keep existing code using the compatibility while start to tapping on more MZ capabilities via its native API.

The new `mz_stream_get_interface()` and the call to `mz_stream_mem_set_grow_size()` would still be necessary (well, the latter can be dropped, but it'd be nice to keep it for more ideal memory use, the default 4K buffer seems quite small for this use-case.)

nmoinvaz:
I am reluctant to add this functionality because I think it is a deeper integration that is more that just trying to maintain compatibility. I think at this point people should just be using the new API. I did however, add one of the functions references in the diff.

vszakats:
Thanks, that API is useful, though unless the MZ and stream objects are exposed as well, one can't implement this or other similar functionality without maintaining a local patch for the minizip source code. The other missing bit is tweaking the newly created stream object inside `zipOpen*()` (or `unzOpen*()`), which isn't currently possible without patching minizip itself. This particular case could be mitigated by bumping up the default buffer size of memory streams to avoid memory fragmentation, assuming that most "typical" .zip files cannot fit into 4 KB (the current buffer grow-size). 32K or 64K may be a good compromise here.

I'm at this point not able to drop the old minizip1 API for these reasons:
- It has been around for a very long time, so it's well-known and stable.
- There is substantial amount of existing (often fragile) code that uses the old API. Rewriting this code to use the MZ API is as large amount of work and risk.
- I'm using an open-source wrapper for minizip functionality, which wrapper still offers compatibility with minizip1. This is necessary because this very minizip fork (or rather: re-implementation):
  - isn't available via most package managers, so the only option is to vendor it.
  - doesn't yet work on all platforms, e.g. Linuxes (in part due to `<bsd/stdlib.h>` requirement). (that downstream open-source project (written in pure ANSI C89) offers a very wide range of portability options, and it's important to have a minizip subsystem in all of them.)
- The MZ API/ABI is still in flux and changes often, even in minor version number updates, so one will need to be prepared to keep tweaking the interface code version to version.

It means it's too early to jump to use MZ API directly. But, for optional functionality that is only available via MZ API, it's a fine solution, and this would be greatly aided by allowing to intermix the compatibility functions with the native API. For this, one needs to be able to retrieve the stream object and the MZ object from the compatibility object.

The particular feature implemented by this PR is useful to create .zip files in memory for sending them down the wire to a client (without touching the server disk and using temporary files). That's the scenario I'm using it for. Since I've implemented this already, I'm going to maintain it as a local patch anyway, but maybe it can be useful for others as well. Having said that I understand why maintaining MZ specific extensions to the compatibility isn't desired, but there is just no other option at the moment, unless one can fully drop the compatibility API.

vszakats:
BTW, why dropping the `const` keywords? These are useful to avoid the caller to fiddle with the returned value and also gives ways the C compiler to do optimizations.

nmoinvaz:
I see what you are saying about the mem grow size. I have changed it. The inclusion of <bsd/stdlib.h> is necessary for encryption function arc4random_buf. I made some changes and it is possible to compile without it. I have also added the functions to get the stream that you requested.

vszakats:
Great updates! I like your version much better than this PR. I've already updated and tested the new code and everything works perfect.

I've seen you added `unzGetStream()`. This reminded me the interesting other side of memory stream support — to be able to _unzip_ memory streams. This would need a little bit of different approach and allow to _assign_ an already created stream with the input buffer/size configured. Something alike `unz = unzOpenStream( stream );`.

Here's a patch implementing it. Let me know what you think:
```diff
diff --git a/contrib/hbmzip/3rd/minizip2/mz_compat.c b/contrib/hbmzip/3rd/minizip2/mz_compat.c
index 2343eb8baf..7e079b8a1c 100644
--- a/contrib/hbmzip/3rd/minizip2/mz_compat.c
+++ b/contrib/hbmzip/3rd/minizip2/mz_compat.c
@@ -337,9 +337,7 @@ extern unzFile ZEXPORT unzOpen2(const char *path, zlib_filefunc_def *pzlib_filef
 
 extern unzFile ZEXPORT unzOpen2_64(const void *path, zlib_filefunc64_def *pzlib_filefunc_def)
 {
-    mz_compat *compat = NULL;
     int32_t mode = MZ_OPEN_MODE_READ;
-    void *handle = NULL;
     void *stream = NULL;
 
     if (pzlib_filefunc_def)
@@ -359,6 +357,18 @@ extern unzFile ZEXPORT unzOpen2_64(const void *path, zlib_filefunc64_def *pzlib_
         return NULL;
     }
 
+    return unzOpenStream(stream);
+}
+
+extern unzFile ZEXPORT unzOpenStream(void *stream)
+{
+    mz_compat *compat = NULL;
+    int32_t mode = MZ_OPEN_MODE_READ;
+    void *handle = NULL;
+
+    if (stream == NULL)
+        return NULL;
+
     handle = mz_zip_open(stream, mode);
 
     if (handle == NULL)
diff --git a/contrib/hbmzip/3rd/minizip2/mz_compat.h b/contrib/hbmzip/3rd/minizip2/mz_compat.h
index 36c26ab7a7..7b79ed1ce5 100644
--- a/contrib/hbmzip/3rd/minizip2/mz_compat.h
+++ b/contrib/hbmzip/3rd/minizip2/mz_compat.h
@@ -255,6 +255,7 @@ extern unzFile ZEXPORT unzOpen(const char *path);
 extern unzFile ZEXPORT unzOpen64(const void *path);
 extern unzFile ZEXPORT unzOpen2(const char *path, zlib_filefunc_def *pzlib_filefunc_def);
 extern unzFile ZEXPORT unzOpen2_64(const void *path, zlib_filefunc64_def *pzlib_filefunc_def);
+extern unzFile ZEXPORT unzOpenStream(void *stream);
 
 extern int ZEXPORT unzClose(unzFile file);
 extern int ZEXPORT unzGetGlobalInfo(unzFile file, unz_global_info* pglobal_info32);
```

nmoinvaz:
Ok I have added that function too.

vszakats:
Excellent, thank you! (Also for the fixes to my patch)

PR #288, opened 2018-07-26
Read ZIP global file comment right after reading EOCD fields.

### LINKED ISSUE #287: Global comment is not properly read
When trying to access the global comment of a ZIP file, garbage is returned. The problem seems to be at the `mz_zip_read_cd` function in mz_zip.c. After reading all of the End of Central Directory fields the stream pointer is moved (because there's an attempt to find the ZIP64 EOCD and/or validate the first CD record) and only then the global comment is read, but the stream pointer is already anywhere else.

juanii:
Fixes #287

### DIFF
--- a/mz_zip.c
+++ b/mz_zip.c
@@ -99,7 +99,7 @@ typedef struct mz_zip_s
 
 /***************************************************************************/
 
-// Locate the central directory of a zip file (at the end, just before the global comment)
+// Locate the end of central directory of a zip file (at the end, just before the global comment)
 static int32_t mz_zip_search_eocd(void *stream, uint64_t *central_pos)
 {
     uint8_t buf[1024 + 4];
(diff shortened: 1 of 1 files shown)


nmoinvaz:
Are you sure this fixes it? It looks like it is just changing some comments.

nmoinvaz:
I think my commit should fix it. Please take a look.

juanii:
Oh god! I made the change for this PR and fixed those comments, then staged the whole file and _un_staged the wrong lines!

There's a catch with your commit: if the file happens to be a ZIP64, the `if` statement in line 259 will be true and inside its body a call to `mz_zip_search_zip64_eocd` is made, which again moves the stream pointer. My fix moved the same code block as yours, but right after the global comment length is read, in line 257.

I'm not very experienced in gitbhub and I'm not sure how this is fixed: should I create a new PR? can I remove the previous commit from this PR and add the correct one? is it easier for you to make the change?

Sorry for my poor attention on this one.

nmoinvaz:
It would probably be easier for me to make it. Take a look at my change if it is what you mean.

juanii:
That's correct!

Thanks and sorry again for the sloppy PR

nmoinvaz:
I made the comment changes too since your original pull request was against master branch instead of dev branch.

nmoinvaz:
Thanks for your fixes!

PR #290, opened 2018-07-27
fix zipOpenNewFileInZip*() to allow passing level in raw mode

vszakats:
This syncs this behaviour with original minizip (as of 1.2.11)

Ref: https://github.com/nmoinvaz/minizip/issues/289#issuecomment-408515522


vszakats:
(Wrong branch, sorry)

PR #296, opened 2018-08-16
adds option to link against system bzip2 library

UncleAlbie:
I'm packaging this project for Fedora as it should eventually replace the minizip from the zlib project. I would need to be able to link against the system bzip2 library to comply with our guidelines.

These changes don't change the default behavior since the option is set to OFF by default.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,6 +13,7 @@ option(USE_AES "Enables AES encryption" ON)
 option(COMPRESS_ONLY "Only support compression" OFF)
 option(DECOMPRESS_ONLY "Only support decompression" OFF)
 option(BUILD_TEST "Builds minizip test executable" OFF)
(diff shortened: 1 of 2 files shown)


nmoinvaz:
I think instead of having a USE_SYSTEM_BZIP, that it should check to see if BZIP package exists on the system and if it does then default to that, otherwise it would use the one supplied in the source directory.

UncleAlbie:
I think you're right. That seems to be a better solution. I will implement this for lzma and aes as well in a new PR. I'll close this one now. Thanks.

nmoinvaz:
@UncleAlbie - It is not possible to do it with LZMA since there are changes that had to be made to LZMA library in order to support zip format. Do any distros come with the same aes library?

UncleAlbie:
I had to check up about aes and it seems like they don't come with this version. So it makes sense only with the bzip2 library. Thank you for clarification.

PR #297, opened 2018-08-17
use system Bzip2 library if available

UncleAlbie:
Checks if BZip2 installation is present and link against the system one if so.

Previous discussion: #296


nmoinvaz:
Looks good. Nice work. Thank you!

PR #301, opened 2018-08-28
README: Update/add Travis CI links, use HTTPS

vszakats:
* also fold some long lines
* remove double EOL at EOF


nmoinvaz:
Thanks!

PR #306, opened 2018-08-31
mz_os_posix.c: remove duplicate #include

vszakats:


nmoinvaz:
I believe stdlib.h is used for malloc/free in that file.

PR #307, opened 2018-09-03
Fix infinite loop if a read operation does not return any bytes

driquet:
While I was fuzzing a program of mine, I encountered an infinite loop. The input file was obviously malformed, and minizip was trying to copy N bytes from a stream with M bytes (N > M). In that case, the limit is reached and `mz_stream_mem_read` does not return any bytes. That causes an infinite loop in `mz_stream_copy`.


nmoinvaz:
Thanks

PR #308, opened 2018-09-04
enable warnings for gcc/clang, misc minor fixes

vszakats:
* enable `-W -Wall` for gcc and clang
* fix `strncpy()` calls that may potentially leave the target buffer without
  a terminating zero byte (in `mz_zip_rw.c`)
* fix warnings
* remove `const` from `mz_path_remove_filename()` parameter,
  the function is writing the passed buffer
* fix pedantic warnings (in `mz_strm_buf.c`):
  `warning: suggest braces around empty body in an ‘if’ statement [-Wempty-body]`
* fix spaces at EOL
* fix a redefinition warning for `HAVE_ARC4RANDOM_BUF` on macOS

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -111,6 +111,10 @@ if("${CMAKE_SYSTEM_NAME}" STREQUAL "WindowsStore")
     add_definitions(-DMZ_WINRT_API)
 endif()
 
+if(CMAKE_COMPILER_IS_GNUCC OR CMAKE_COMPILER_IS_CLANG)
+    add_compile_options(-W -Wall)
+endif()
+
 if(UNIX)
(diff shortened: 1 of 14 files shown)


nmoinvaz:
Thanks for the pull request. I’m currently away and it will be some time before I can take a look at it since there are a lot of changes. Thanks!

vszakats:
@nmoinvaz: No worries! To make the patch easier to read (and significantly shorter), it's useful to hide whitespace changes.

vszakats:
Fixed rest of `strncpy()` calls, mainly those where target was a dynamic buffer, so C compiler didn't show a warning.

nmoinvaz:
Thanks!

PR #310, opened 2018-09-07
Fix minor typo

Coeur:


nmoinvaz:
string_method can only be 6 characters long. %6s%c

Coeur:
%6s means there will be minimum 6 characters printed, eventually padded with spaces at the beginning. But I get it may not look nice if one line is longer than the others, so I'll not touch string_method.

PR #315, opened 2018-09-18
Add mz_zip_[reader,writer]_set_progress_interval

chrfle:
With mz_zip_reader_set_progress_interval the user can specify how
often in milliseconds the progress callback should be called. This
makes it possible to make a smoother progress bar when using
minizip from a GUI application.

### DIFF
--- a/mz_os_posix.c
+++ b/mz_os_posix.c
@@ -33,6 +33,11 @@
 #  include <stdlib.h>
 #endif
 
+#ifdef __APPLE__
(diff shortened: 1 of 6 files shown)


nmoinvaz:
Please make changes against _dev_ branch next time.

chrfle:
Certainly! If I get time for it (unlikely), I would like to see if cmake can detect clock_gettime, as it should be present from Sierra.

PR #316, opened 2018-09-19
mz_zip_reader_entry_read: remove const from buf parameter

vszakats:


vszakats:
(wrong branch, sorry)

PR #323, opened 2018-10-08
Avoid double free on comment

driquet:
This will prevent the zip context to free multiple times the comment value, if for some reason the close function is called several times.


nmoinvaz:
Thanks!

Issue #201, opened 2017-10-31
Why 64k limitation when read uncompressed file?


Jiaoweiquan:
https://github.com/nmoinvaz/minizip/blob/db8df5bd96fa372f3b8240cc55f8fca2937aa15c/src/mz_zip.c#L1301

nmoinvaz:
```
typedef struct z_stream_s {
    z_const Bytef *next_in;     /* next input byte */
    uInt     avail_in;  /* number of bytes available at next_in */
    uLong    total_in;  /* total number of input bytes read so far */

    Bytef    *next_out; /* next output byte will go here */
    uInt     avail_out; /* remaining free space at next_out */
    uLong    total_out; /* total number of bytes output so far */

    z_const char *msg;  /* last error message, NULL if no error */
    struct internal_state FAR *state; /* not visible by applications */

    alloc_func zalloc;  /* used to allocate the internal state */
    free_func  zfree;   /* used to free the internal state */
    voidpf     opaque;  /* private data object passed to zalloc and zfree */

    int     data_type;  /* best guess about the data type: binary or text
                           for deflate, or the decoding state for inflate */
    uLong   adler;      /* Adler-32 or CRC-32 value of the uncompressed data */
    uLong   reserved;   /* reserved for future use */
} z_stream;

typedef unsigned int   uInt;  /* 16 bits or more */
```

uInt is only guaranteed to be 16 bits.

evanmiller:
I opened a duplicate issue #227. I would suggest checking the size of `uInt` on the local system - this limitation seriously undermines the usefulness of minizip on modern computers.

Issue #91, opened 2017-02-27
Extracting uncompressed file from ZIP produces endless loop


zett42:
Extracting uncompressed file with size < 65536 is OK.
Extracting uncompressed file with size == 65536 results in extracted file with size of 0.
Extracting uncompressed file with size >= 65537 produces an endless loop.

Can be reproduced with minizip master branch and current zlib version 1.2.11.

Test files:
[test65536.zip](https://github.com/nmoinvaz/minizip/files/803729/test65536.zip)
[test65537.zip](https://github.com/nmoinvaz/minizip/files/803730/test65537.zip)

nmoinvaz:
This has been fixed. Thanks!

zett42:
I can confirm that the issue no longer exists.
Thank you, much appreciated!

puetzk:
Just commenting to link the issue to its commit and add some commentary, since it took me some hunting to find why my program was getting UNZ_PARAMERROR. The related change is fac878383c241f97df2598000cfadc5e4ee1065e, and it's not so much a fix for larger sizes, as forbidding one to even try.

zlib itself doesn't explicitly have such a limitation. The only reason sizes > 65535 don't work is because https://github.com/nmoinvaz/minizip/blob/5b7b3e5c4bf1ce8fd4d3da2057760972832a972f/unzip.c#L1305 is casting the avail_out value down to uint16_t. But I guess it's just being defensive; zlib's `uInt` is just `unsigned int`, and ANSI C89 would legally allow that to be a type with only [0,65535], even though most compilers nowadays would support up to 4GiB.

But on some old or unusual platform this defensiveness could be needed.

nmoinvaz:
If somebody wants to submit a pull request against the 1.2 branch I will accept it.

puetzk:
Yeah, I need to modernize this caller more someday, but the API changed a lot.

Why the complex check involving if (UINT_MAX == UINT16_MAX), instead of just if len > UINT_MAX? The latter would seem to express zlib's limitations on its own if the casts in minizip are gone...

nmoinvaz:
It is a quick check to see if uint on the platform is 16 bit. 

Here is some backstory:

https://github.com/nmoinvaz/minizip/issues/201
https://github.com/nmoinvaz/minizip/pull/228/commits/3064a0e5519ddd360d104b45964c734b2b5ec5cb

puetzk:
> It is a quick check to see if uint on the platform is 16 bit.

[...]

PR #327, opened 2018-10-27
Allow reading > 64k at a time on platforms where zlib supports this

puetzk:
zlib uses uInt (`unsigned int`) for avail_in and avail_out.
This type can be limited to 64k and still conform to ANSI C,
but on most platformsit's bigger.

Remove some unnecessarily-narrow casting and only enforce
the actual limit (UINT_MAX), which might be 64k or might be larger.

This is per the recent discussion at the end of #91 and is essentially a version of the fix for #201 (3064a0e5519ddd360d104b45964c734b2b5ec5cb and 930f9188a5fb39947812d21306c6ae587fbe7c66) for the 1.2 branch

### DIFF
--- a/unzip.c
+++ b/unzip.c
@@ -1298,16 +1298,17 @@ extern int ZEXPORT unzReadCurrentFile(unzFile file, voidp buf, uint32_t len)
         return UNZ_END_OF_LIST_OF_FILE;
     if (len == 0)
         return 0;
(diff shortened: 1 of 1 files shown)


puetzk:
I have verified that test65536.zip and test65537.zip from #91 extract using miniunz_exec.exe, as well as checking the case with my own data where I hit this.

Issue #328, opened 2018-10-30
want legacy encoding support cjk dbcs codepage


xwxbug:
in 90's ,many file use dbcs character sets in chinese japanese korea(CJK).or other dbcs  character sets.
minizip support cp437 codepage ,need a option to support cp932 cp936 cp950 and so on.

[test_demo_cp950.zip](https://github.com/nmoinvaz/minizip/files/2527729/test_demo_cp950.zip)

nmoinvaz:
I currently don't have any plans to support additional character sets - only CP437 is named in the ZIP app note. If you make a pull request I will consider it. Thanks.

PR #330, opened 2018-10-30
support dbcs character sets at win32

xwxbug:
sorry,i use win32...so.code only works on win32.

for #328


nmoinvaz:
I have taken what you done and incorporate it in a different way. See the dev branch. Thank you.

PR #343, opened 2018-11-08
Cast day/month/year elements to uint32_t

akmistry:
If the sum of day/month/year elements exceeds 32767, ubsan will tag it as undefined behaviour since the tm struct elements are int which is only defined to be at least 16-bits and signed. Casting these elements to uint32_t guarantees the calculations will be defined.


akmistry:
Sorry, I'm still trying to figure out github's confusing UI.

nmoinvaz:
Looks good. Thanks!

PR #345, opened 2018-11-14
make inttypes.h optional

vszakats:
`inttypes.h` isn't available in certain environments, but instead of relying on CMake to create a fake local copy, let's handle the matter in the source code, and let it work with any build system. If `HAVE_INTTYPES_H` is defined, `inttypes.h` will be used, otherwise the required values will be defined locally.


akmistry:
Can the condition be inverted (i.e. have NO_INTTYPES_H_ instead)? I think the default should be to assume a sane environment, and have the define available for those that need it.

vszakats:
Speaking of toolchain features, for me the `NO_INTTYPES_H` method would look rather unusual, so I'd prefer to stick with the more familiar way. I also find it better to assume the _least_ by default. (Original minizip also didn't require `inttypes.h`.)

nmoinvaz:
I’m on vacation so I will take a look when I get back.

nmoinvaz:
Sorry I did not merge your changes. I wasn't sure at the beginning I was going to do it so similar as your changes.

PR #350, opened 2018-11-28
Use absolute paths for install directories.

UncleAlbie:
Absolute paths should be used in pkgconfig file.


nmoinvaz:
I think it might make more sense to use CMAKE_INSTALL_FULL_* because if CMAKE_INSTALL_PREFIX is empty it will try to install into root / instead of bin of the current directory.

UncleAlbie:
I think you're right. I didn't think of that.

I've edited this PR accordingly.

PR #357, opened 2019-01-11
Added prefix to CMakeLists.txt's options

SleepySpaceDad:
Added prefixes to the options to make using minizip as an embedded dependency easier


nmoinvaz:
I think it might be better to do "MZ_" instead of "MINIZIP_". Also maybe the travis yaml needs to be updated with the correct names and that is why it is failing..

PR #365, opened 2019-02-13
Add correct shared library prefix

UncleAlbie:
Shared library prefix is missing since the project has been renamed. That causes issues on UNIX systems.


nmoinvaz:
It looks like this is failing on windows.

nmoinvaz:
I have removed setting the PREFIX altogether. Hopefully it fixes your issue.

PR #366, opened 2019-02-18
fixed issue where install target failed due to the fact that the .pc …

mschmieder:
…was never generated

Hey @nmoinvaz! I like what you did out of my first CMake configuration :)
I think a slight "glitch" with the CMake install target and the package conf file found it's way in your last releases.

At least on my side I'm not able to install successfully anymore due to the fact, that the `minizip.pc.cmakein` is search at the wrong place.

Please check this simple patch and keep up the great work!


nmoinvaz:
I'm glad you like all the improvements. Thanks for the patch!

PR #367, opened 2019-02-19
Fix package config install target

mschmieder:
@nmoinvaz sorry, but I screwed up the last fix - it actually did not fix the underlying issue

I removed the complete `if...end` statement since this is not necessary, CMake will take care of the generation. 
To make it up to you, I also removed the unnecessary trailing spaces :)

Please consider upgrading your travis CI to also do the installation step in order to figure these kinds of problems out in the first place.


nmoinvaz:
I was wondering about that EXISTS statement because I saw zlib didn't have one. But I didn't think it hurt anything. Thanks again for the fixes!

nmoinvaz:
Actually i think it should be: `if (EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/minizip.pc.cmakein)` where it would only configure the file if it exists in the source tree.

mschmieder:
I wouldn’t do that - why wouldn’t the file be there? It’s part of your repository as all the other sources are, right? So it’s a „fact“ and nothing „variable/condionally generated“ 

Also the current source dir reference is not needed since the relative path is easier to read and maintain.

PR #370, opened 2019-03-07
Allow recursively creating Windows path to fix #369

### LINKED ISSUE #369: Cannot create dir from full Windows path
This seems to be a bug.
When extracting an archive, directory creation fails when a subdirectory is to be extracted, but its parent directory in the archive has not yet been created. On the below exawmple structue, this issue occurs when trying to extract file_x earlier than file_y.

Example - extracting the below archive into **d:\temp**:
/dir1/
  /subdir1/
  /subdir1/file_x
  /file_y
/dir2/

What specifically happens is that as the **mz_dir_make** function in **mz_os.c** runs to create the directory **dir1/subdir1** for **file_x** on the file system, it goes ahead and breaks down the full Windows path **d:\temp\dir1\subdir1** to components to be able to create it level-by-level, and it breaks it up based on path-separators (obviously), attempting to create them in this order:
d:\
d:\temp
d:\temp\dir1
d:\temp\dir1\subdir1

That's to logic at least - but the first iteration fails, as **CreateDirectoryW** in **mz_os_win32.c** yields **ERROR_ACCESS_DENIED** for a drive letter, preventing the loop from continuing with the iteration on the target path, and ends up returning **MZ_INTERNAL_ERROR** from **mz_zip_reader_save_all**.

This is specific to Windows and to archives where subdirs occur before files on the same level inside a directory.

tamasszarka:
Please see issue here: https://github.com/nmoinvaz/minizip/issues/369

### DIFF
--- a/mz_os_win32.c
+++ b/mz_os_win32.c
@@ -357,8 +357,8 @@ int32_t mz_os_make_dir(const char *path)
 
     if (CreateDirectoryW(path_wide, NULL) == 0)
     {
-        // skip the drive letter when creating a directory recursively
-        if (wcslen(path_wide) == 2 && path_wide[1] == L':')
+        /* Skip the drive letter when creating a directory recursively */
+        if ((wcslen(path_wide) == 2) && (path_wide[1] == L':'))
             err = MZ_OK;
 
         if (GetLastError() != ERROR_ALREADY_EXISTS)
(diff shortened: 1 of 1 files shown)


nmoinvaz:
Does using wsclen require a additional #include or does <windows.h> cover it?

PR #387, opened 2019-05-01
Unzipping password protected files with no password

Coeur:
When a password-protected archive doesn't use password for its content, it shouldn't error out.

Here is a sample file
[EmptyWithParentDirectory.zip](https://github.com/nmoinvaz/minizip/files/3134399/EmptyWithParentDirectory.zip)

(taken from unit test `testZippingAndUnzippingEmptyDirectoryWithPassword` in temporary branch https://github.com/ZipArchive/ZipArchive/tree/minizip286)


nmoinvaz:
Looks good. Thanks!

PR #388, opened 2019-05-01
Fix support for creating empty archives

Coeur:
The creation of empty archives (with 0 entries) got broken by https://github.com/nmoinvaz/minizip/commit/64faa259d5acc39b842b2eab4eb5b1568acf3ce0.

We do want a Central Directory, otherwise the file created would be 0 bytes and not a valid zip file.


nmoinvaz:
Thanks! I accidentally created a revert pull request.

PR #390, opened 2019-05-02
Gitignore for cmake generated folders

Coeur:
This won't automatically ignore the few generated files at the root of the repo, like `CMakeCache.txt`, but it will help having more visibility on modified source files by ignoring the generated folders.

I've also added `.idea/` (used by AppCode and other JetBrains editors), but if you prefer to stick to cmake stuff, we can remove `.idea/` from the list.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,9 @@
+# cmake generated folders
+Debug/
+CMakeFiles/
+CMakeScripts/
+minizip.build/
+minizip.xcodeproj/
+
+# JetBrains editors
+.idea/


nmoinvaz:
I hesitate to do this because CMake generates solutions and such for whatever generator you use. So we'd end up doing all these git ignores for the various solution types CMake supports.

Coeur:
OK. But do you know if it's possible to define in CMakeLists.txt a parent folder for all those generated content? If that was possible, then there would be only one folder to ignore.

nmoinvaz:
I think it is possible to build/generate in a separate directory but most of the time I do not do it.

Coeur:
Anyway, I can deal with it, so I'm closing this PR for now.

PR #391, opened 2019-05-02
Allowing compilation on iOS/tvOS when MZ_ZIP_NO_SIGNING is defined

Coeur:
[CMSDecoder](https://developer.apple.com/documentation/security/cmsdecoder) is only on macOS, not on iOS or tvOS.


nmoinvaz:
Thanks!. I have applied the same changes to the rest of the crypt interfaces.

PR #393, opened 2019-05-06
Podspec update

Coeur:
This updates the podspec for the new files (mz.h, mz_os_posix.c, etc.).
This adds a job in travis to validate that the default_subspecs are installable.

### DIFF
--- a/Minizip.podspec
+++ b/Minizip.podspec
@@ -13,41 +13,46 @@ DESC
   s.homepage = 'https://github.com/nmoinvaz/minizip'
   s.authors = 'Nathan Moinvaziri', 'Gilles Vollant'
 
-  s.source   = { :git => 'https://github.com/nmoinvaz/minizip.git' }
+  s.source   = { :git => 'https://github.com/nmoinvaz/minizip.git', :tag => "#{s.version}" }
   s.libraries = 'z', 'iconv'
-  s.default_subspecs = 'Core', 'PKWARE', 'crypt_apple', 'BZIP2'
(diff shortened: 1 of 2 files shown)


Minizip.podspec
```diff
@@ -14,24 +14,42 @@ DESC
...
 
-  s.subspec 'AES' do |sp|
+  s.subspec 'PKWARE' do |sp|
     sp.dependency 'Minizip/Core'
-    sp.source_files = 'lib/aes/*.{c,h}', 'mz_strm_aes.{c,h}'
+    sp.source_files = 'mz_strm_pkcrypt.{c,h}'
+    sp.pod_target_xcconfig = { 'GCC_PREPROCESSOR_DEFINITIONS' => 'HAVE_PKCRYPT' }
+  end
+
+  s.subspec 'crypt_apple' do |sp|
```
nmoinvaz: I think I'd perfer the names of the submodules to closely match the names of the CMAKE options. Maybe something like this, if it is doable? 

PKWARE -> 'PKCRYPT'
crypt_apple -> 'WZAES - CommonCrypto'
crypt_brg -> 'WZAES - BRG'

What do you think?

nmoinvaz:
With the travis integration, I think it is a good idea, but I worry about being able to maintain it in the future because I know nothing about podspec.

Minizip.podspec
Coeur: OK for WZAES_* and PKCRYPT. But for Core, I'll leave it capitalized, as we normally don't uppercase spec names anyway.
Don't worry for maintaining, when people will use it (maybe ZipArchive), we'll pull request if needed.

PR #396, opened 2019-05-15
Fix to compile with mingw-w64

zedxxx:


Coeur:
Just curious, if `_MSC_VER` is not defined, then `if (_MSC_VER < 1900)` two lines below will be a problem, isn't it?

zedxxx:
It compiles without errors for me. I used latest builds mingw-w64 from msys2 (32&64 bit versions) for test.

zedxxx:
Will this way be more correct?
```
#if defined(_MSC_VER) || defined(__MINGW32__)
#  define localtime_r(t1,t2) (localtime_s(t2,t1) == 0 ? t1 : NULL)
#  if defined(_MSC_VER) && (_MSC_VER < 1900)
#    define snprintf _snprintf
#  endif 
#endif
```

nmoinvaz:
@zedxxx does this compile for you?

```
#if defined(_MSC_VER) || defined(__MINGW32__)
#  define localtime_r(t1,t2) (localtime_s(t2,t1) == 0 ? t1 : NULL)
#endif
#if defined(_MSC_VER) && (_MSC_VER < 1900)
#  define snprintf _snprintf
#endif
```

zedxxx:
Yes.

nmoinvaz:
Ok, if you update the pull request I can merge it in.

nmoinvaz:
Thanks!

PR #402, opened 2019-06-19
Fix return of mz_stream_zlib_get_prop_int64() for PROP_COMPRESS_WINDOW

plo-:
There's a missing `break;` in the switch/case of `mz_stream_zlib_get_prop_int64()` when prop is `MZ_STREAM_PROP_COMPRESS_WINDOW`, resulting in a `MZ_EXIST_ERROR` return instead of `MZ_OK`.


nmoinvaz:
Thank you!

PR #406, opened 2019-07-05
Fix BUILD_BYPRODUCTS argument for zlib

axnsan12:
When using the Ninja generator, CMake relies on the BUILD_BYPRODUCTS argument of ExternalProject_Add to determine how to build libz.a. 

Previously this was wrongly set to the empty variable `ZLIB_STATIC_LIBRARIES` instead of `ZLIB_LIBRARIES`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -80,8 +80,12 @@ if(MZ_ZLIB AND NOT MZ_LIBCOMP)
         if(WIN32)
             set(ZLIB_LIBRARIES
                 ${ZLIB_INSTALL_DIR}/lib/zlibstatic$<$<CONFIG:Debug>:d>.lib)
+            set(ZLIB_BYPRODUCTS
+                ${ZLIB_INSTALL_DIR}/lib/zlibstatic.lib ${ZLIB_INSTALL_DIR}/lib/zlibstaticd.lib)
         else()
             set(ZLIB_LIBRARIES
+                ${ZLIB_INSTALL_DIR}/lib/libz.a)
+            set(ZLIB_BYPRODUCTS
                 ${ZLIB_INSTALL_DIR}/lib/libz.a)
         endif()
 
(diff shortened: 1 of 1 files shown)


Coeur:
CI build failed with:
```
-- Looking for stdint.h
CMake Error at C:/Program Files/CMake/share/cmake-3.12/Modules/ExternalProject.cmake:2118 (add_custom_command):
  add_custom_command called with OUTPUT containing a "<".  This character is
  not allowed.
Call Stack (most recent call first):
  C:/Program Files/CMake/share/cmake-3.12/Modules/ExternalProject.cmake:2925 (ExternalProject_Add_Step)
  C:/Program Files/CMake/share/cmake-3.12/Modules/ExternalProject.cmake:3104 (_ep_add_build_command)
  CMakeLists.txt:106 (ExternalProject_Add)
```

axnsan12:
Hmm, right... That seems to be because of the generator expression at line 82:

`${ZLIB_INSTALL_DIR}/lib/zlibstatic$<$<CONFIG:Debug>:d>.lib`

which fails since the BYPRODUCTS option needs to know its value at configure time, it seems. I'm not really sure what the most elegant fix would be here...

nmoinvaz:
Maybe the key is just to list both zlibstatic.lib and zlibstaticd.lib as byproducts for WIN32 regardless of what configuration is being used.

axnsan12:
From my tests, this does work as expected. A build rule is generated for both files, but only one of them is required at any one time, so build result caching works as expected.

axnsan12:
Maybe we should add some Travis tests using Ninja? Or arbitrarily switch some of the existing ones to use Ninja?

nmoinvaz:
I would be open to adding Ninja generator. You mentioned `ZLIB_STATIC_LIBRARIES`, where does come from?

axnsan12:
No idea, I guess it was left as a typo or a refactoring error - on the BYPRODUCTS line of the ExternalProject_Add command.

nmoinvaz:
Are you talking about an older version of the library it existed in?

axnsan12:
Yeah, that was my assumption, but it seems that line is the same since it was added: https://github.com/nmoinvaz/minizip/commit/8cc08008767c2f274ca946e871ea79eed3bcd73e

So I guess building zlib never worked when using Ninja

nmoinvaz:
I think it makes more sense if you rename `ZLIB_STATIC_LIBRARIES` to `ZLIB_BYPRODUCTS`. Then I can merge the PR.

PR #411, opened 2019-08-10
Use memcpy instead of strncpy

MangaD:
1. `memcpy` is faster than `strncpy`
2. `strncpy` was depending on the size of the source string, therefore useless in comparison with `strcpy`
3. I was getting a compiler warning for `strncpy` for the above reason.

### DIFF
--- a/aes/entropy.c
+++ b/aes/entropy.c
@@ -3,6 +3,7 @@
 #else
 #include <sys/stat.h>
 #include <fcntl.h>
+#include <unistd.h>
 #endif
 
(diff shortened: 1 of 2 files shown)


Coeur:
Your branch is a fork from June 2015 (more than 4 years ago), without retrieving any update since then: https://github.com/MangaD/minizip/commits/master

In my opinion, there is no way that your pull request can be accepted as is.

MangaD:
You're absolutely right, I didn't notice I was dealing with a forked repository. Closing this now.

PR #415, opened 2019-09-06
Fix issue with truncated ZLIB output.

### LINKED ISSUE #414: CRC issue with deflate compression and a read loop
With a particular output buffer, a zip file with an AES encrypted (this may be red herring) and zlib compressed item fails reading with a CRC error.  I noticed it with a buffer size of 8192 and a uncompressed item of size  7077903, but I saw it at some other combinations. 

On investigation, it appears that zlib::inflate may read data from the input buffer and/or write it to the output buffer, but not neccessarily both. "[If inflate returns Z_OK and with zero avail_out, it must be called again after making room in the output buffer because there might be more output pending.](https://www.zlib.net/manual.html)"  This means: zlib may consume X bytes, decompress them to an internal buffer, and then output the decompressed results over multiple calls.  In the current implementation of `mz_strm_zlib`, the decompress loop will be escaped if there is no data to read and the minizip controlled buffer is empty, though there may be more data to output.

[main.cpp.zip](https://github.com/nmoinvaz/minizip/files/3581666/main.cpp.zip) is a test script that shows the problem.

Tested against minizip 2.8.9 (also 2.8.1) with
- iconv 1.15
- OpenSSL 1.0.2p
- bzip2 1.0.6
- zlib 1.2.11

rddesmond:
Fixes #414 in zlib by continuing to loop until no work is done, an error is produced, or the output buffer is full.  (Previously, it would also break if no input was provided, but zlib can continue to output as needed if the output buffer is too small to take everything.)

### DIFF
--- a/mz_strm_zlib.c
+++ b/mz_strm_zlib.c
@@ -168,8 +168,6 @@ int32_t mz_stream_zlib_read(void *stream, void *buf, int32_t size)
 
             if (read < 0)
                 return read;
-            if (read == 0)
-                break;
 
             zlib->zstream.next_in = zlib->buffer;
             zlib->zstream.avail_in = read;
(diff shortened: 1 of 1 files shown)


nmoinvaz:
Thank you.

PR #419, opened 2019-09-15
Added mz_zip_rw.*, mz_strm_split.* files to podspec. Use system bzip2…

ShvetsovAndrey:
… for subspec BZIP2.

### DIFF
--- a/Minizip.podspec
+++ b/Minizip.podspec
@@ -51,7 +51,7 @@ DESC
     sp.pod_target_xcconfig = { 'GCC_PREPROCESSOR_DEFINITIONS' => 'HAVE_BZIP2' }
   end
   
-  s.subspec 'BZIP2_LIB' do |sp|
+  s.subspec 'BZIP2_SYSTEM' do |sp|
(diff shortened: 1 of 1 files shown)


nmoinvaz:
@Coeur thoughts?

nmoinvaz:
I think it might be better to have a `BZIP2` and a `BZIP2_LIB`. So BZIP2 would link the bzip2 files in the minizip repo and BZIP2_LIB would link against the system bz library.

nmoinvaz:
Thanks!

Coeur:
So, a bit late for the review, but Nathan is right, having two distinct subspecs here is better. In the future, we may change the default_subspecs to use BZIP2_SYSTEM if it works on all platforms (macOS, iOS, ...), but that is something that I haven't analyzed at all.

Regarding mz_zip_rw.* and mz_strm_spli.*, good catch, as I'm definitely using those: https://github.com/ZipArchive/ZipArchive/tree/master/SSZipArchive/minizip

PR #434, opened 2019-10-23
Unused variable 'central_header_pos'

Coeur:


Coeur:
I'm afraid a single sponsorship will only buy you a coffee or a beer. I would be surprised if it's a viable solution for a living.

nmoinvaz:
Haha well I will enjoy that beer 🍺!

PR #437, opened 2019-12-10
Missing mz_strm_os.h in the podspec

kdev0:
- added a missing mz_strm_os header


kdev0:
> Note that for a podspec, you don't need to have a strict regex match, it would have also work with `{mz_strm_os,mz_strm_os_posix}.{c,h}`

Controversial moment. I prefer that all sources be strictly defined, which excludes the inclusion of unnecessary files in the project. But I updated my last commit according to your style.

PR #448, opened 2020-02-09
mz_os_win32.c: kernel32.dll is always loaded, so use GetModuleHandleW().

GerbilSoft:
The kernel32.dll pathname code had two issues:

1. Using sizeof() for wchar_t[]. sizeof() returns the number of bytes, but the wcs*() functions need number of characters, so we end up telling wcsncat() that the buffer is twice as big as it actually is.

2. wcsncat()'s size parameter is supposed to be the *remaining* size in the buffer, not the total size. wcscat_s() should be used instead, since it takes the total size of the buffer.

None of this is actually needed, since kernel32.dll is used by all Windows programs, so it's always loaded in memory.


nmoinvaz:
Thanks!

PR #450, opened 2020-02-12
fix build error with zlib-ng compat

xuchunmei000:
when build zlib-ng with zlib-compat and ZLIB_COMPAT-ON,
error like this:
/home/minizip/minizip-2.9.1/mz_crypt.c:19:14: fatal error:
-ng.h: No
such file or directory
include "zlib-ng.h"
/home/minizip/minizip-2.9.1/mz_strm_zlib.c:19:12: fatal error:
zlib-ng.h: No such file or directory
include "zlib-ng.h"

Signed-off-by: Chunmei Xu <xuchunmei@linux.alibaba.com>

### DIFF
--- a/mz_crypt.c
+++ b/mz_crypt.c
@@ -15,7 +15,7 @@
 
 #if defined(HAVE_ZLIB)
 #  include "zlib.h"
-#  if defined(ZLIBNG_VERNUM)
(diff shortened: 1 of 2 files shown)


nmoinvaz:
Thank you

PR #456, opened 2020-03-10
mz_compat.c: add feature of symlink compression

willson-chen:


willson-chen:
Hi @nmoinvaz , I have no idea about how to remove the error in CodeFactor. Could you please offer some help?

willson-chen:
I am will to add test case to these code blocks. But I found there is no way to test codes in `mz_compat.c` right now. Here are two ways to test them:

* Integrate APIs in `mz_compat.c` to `minizip.c`, which is used to generate command `minizip`. We can add some extra option to specify using APIs in `mz_compat.c` to run test function.
* Considering the command `minizip` is not just for test, we can create a new test function named `minizip_compat`, which is same to `minizip`, but only use APIs in `mz_compat.c`.

@nmoinvaz, how do you think of it? Thanks! :)

Coeur:
@willson-chen when you're modifying `zip_fileinfo` by adding new fields, you're breaking the compatibility with minizip1 api. You can consider all the functions and structs in this file as deprecated.

willson-chen:
Oh, sorry. I knew nothing about minizip1. I should consider the compatibility. Let me close this PR.

PR #469, opened 2020-04-11
Have cmake_build_type propagate to zlib

hsdk123:
Fixes https://github.com/nmoinvaz/minizip/issues/467


nmoinvaz:
I think this will only work for single project generators - not multi project generators like VS correct?

hsdk123:
Just tested on visual studio using the multi config build and switching between Debug, RelWithDebInfo, etc. - seems to work fine with the changes.
```
2>  CUSTOMBUILD : CMake warning :
2>    Manually-specified variables were not used by the project:
2>
2>      CMAKE_BUILD_TYPE
```
Seems it just ignores the cmake_build_type passed in.

> Have you tried upgrading CMake?
To what version? It seems I'm on 3.15

Is there also any specific reason we're using ExternalProject and not FetchContent?

nmoinvaz:
You are right `FetchContent` is a better experience. I have switched CMakeLists.txt to use FetchContent. It will require people to upgrade to cmake 3.11 or use system zlib or custom ZLIB_ROOT.

PR #473, opened 2020-05-05
ignore generated files: zip.h and unzip.h

michael-brade:
Any objections to starting a .gitignore file? `zip.h` and `unzip.h` are generated by CMake no matter from where it is called. It would be nice to not have a changed repository after every cmake change :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+unzip.h
+zip.h


nmoinvaz:
I would prefer not to check-in a .gitignore at this time. People will end up continually try adding to it overtime for their specific IDE files.

PR #476, opened 2020-05-08
Decorate mz_* API with MZ_EXPORT

jimpark:
This change allows the `mz_*` functions to be DLL exported on Windows when building with `-DBUILD_SHARED_LIBS=ON`

### DIFF
--- a/mz_zip.h
+++ b/mz_zip.h
@@ -60,186 +60,186 @@ typedef int32_t (*mz_zip_locate_entry_cb)(void *handle, void *userdata, mz_zip_f
 
 /***************************************************************************/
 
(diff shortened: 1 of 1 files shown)


nmoinvaz:
Why did you close it? Also wouldn't every function need to be exported? What if somebody wanted another functions...

jimpark:
I created a new one since I was missing an include of mz.h.

There's now two ways to achieve this with CMake. Either you choose which API you want exported via something like MZ_EXPORT or set a CMake variable WINDOWS_EXPORT_ALL_SYMBOLS=ON. But if you do the CMake way, you will get virtually everything exported and you have little control. And you may also need to remove the export stuff you have already.

PR #482, opened 2020-05-27
Fix incorrect built-in PRIx64 defines

triplef:
The defines using the "ll" prefix should be used on 64-bit, not 32-bit.

This is only used if `HAVE_INTTYPES_H` is not defined, which is probably why it hasn’t been caught before.

### DIFF
--- a/mz.h
+++ b/mz.h
@@ -194,8 +194,20 @@ typedef unsigned long long uint64_t;
 #ifndef PRId8
 #  define PRId8  "hhd"
 #endif
+#ifndef PRIu8
+#  define PRIu8  "hhu"
+#endif
+#ifndef PRIx8
+#  define PRIx8  "hhx"
+#endif
 #ifndef PRId16
 #  define PRId16 "hd"
+#endif
+#ifndef PRIu16
+#  define PRIu16 "hu"
+#endif
+#ifndef PRIx16
+#  define PRIx16 "hx"
 #endif
 #ifndef PRId32
 #  define PRId32 "d"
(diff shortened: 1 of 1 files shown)


nmoinvaz:
Does this work correctly on Windows? I believe on macOS unsigned long is 64-bit but on Windows unsigned long is 32-bit.

triplef:
Good question. Unfortunately I don’t have a good way to check, but I was wondering if there’s may be a better solution for this check.

I found this code in GCC that is using `sizeof (long) == 8` to check for using the `l` prefix – would that be more reliable than using `ULONG_MAX`?
https://github.com/llvm/llvm-archive/blob/355f21d289664efe10fcc5bd81b3b21b75590209/llvm-gcc-4.2/intl/loadmsgcat.c#L173

nmoinvaz:
@trilef, I think you are right, using `sizeof(long) == 8` is a better check to enable `l` prefix.

triplef:
Cool, I changed it to use `sizeof(long)`.

triplef:
Looks like this is not a good solution either, as this inline conditional doesn’t seem to work when using the defines in a format string as is done here:
https://github.com/nmoinvaz/minizip/blob/master/mz_zip.c#L756

It’ll give me the following error: when using the define using sizeof():
> mz_zip.c:756:68: Called object type 'char [5]' is not a function or function pointer

I think checking `ULONG_MAX` could be the right approach after all, as it should in theory correspond to whether a `long` is 32- or 64-bit and thus whether it should use a `l` or `ll` format prefix.

So with that I’d like to go back to the original solution, but check for `ULONG_MAX == 0xfffffffful` to use the `l` prefix, and otherwise use `ll`. Thoughts?

nmoinvaz:
Ok I think that would be good.

triplef:
Cool. I tested this manually by un-defining `HAVE_INTTYPES_H` and enabling `mz_zip_print`. This also exposed missing defines for 8 and 16 bits, which I added.

nmoinvaz:
Nice! Thank you.

PR #489, opened 2020-06-05
Use lseek64 on Android

### LINKED ISSUE #488: use lseeko64 instead of fseeko on android
see https://android.googlesource.com/platform/bionic/+/master/docs/32-bit-abi.md#32_bit-abi-bugs
`fseeko` variant implementation on android.
may be we can use `lseeko64` on android for better compatibility.

okhowang:
close #488

### DIFF
--- a/mz_strm_os_posix.c
+++ b/mz_strm_os_posix.c
@@ -25,7 +25,10 @@
 /***************************************************************************/
 
 #define fopen64 fopen
-#ifndef MZ_FILE32_API
+#ifdef __ANDROID_API__
+#  include <unistd.h>
+#  define fseeko64(stream, offset, where) (lseek64(fileno(stream), offset, where) >= 0 ? 0 : -1)
+#elif !defined(MZ_FILE32_API)
 #  ifndef NO_FSEEKO
 #    define ftello64 ftello
 #    define fseeko64 fseeko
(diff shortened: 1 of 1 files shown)


nmoinvaz:
I think there should be a check for the android API version number, because it says on newer versions fseeko it is not needed.

nmoinvaz:
Please review my changes to make sure they work for you. You will have to define MZ_FILE32_API in the project.

okhowang:
according to [https://android.googlesource.com/platform/bionic/+/master/docs/32-bit-abi.md#32_bit-abi-bugs](https://android.googlesource.com/platform/bionic/+/master/docs/32-bit-abi.md#32_bit-abi-bugs)
> Android does not require the _LARGEFILE_SOURCE macro to be used to make fseeko and ftello available. Instead they're always available from API level 24 where they were introduced, and never available before then.

firstly, `fseeko` and `ftello` is general available from API level 24.

and `MZ_FILE32_API` seems undocumented. I should set it for platform which has no `ftello` and `fseeko`?

okhowang:
I test with `MZ_FILE32_API` and everthing is ok.
myabe I can just set `MZ_FILE32_API` for android platform.
and this PR can enhance 64bit file ability for android?

nmoinvaz:
I have merged into `dev`  branch. You should be able to set it easily with `-DMZ_FILE32_API=ON`.

PR #494, opened 2020-06-10
Allow libbsd to be explicitly disabled

dwmw2:
When building for a minimal environment like OpenWRT we want to avoid the
dependency on libbsd even if it did happen to be built and available in
the buildroot at the time minizip is compiled.

https://github.com/openwrt/packages/issues/12471


dwmw2:
I don't think those CI failures are related to this change...

PR #497, opened 2020-06-13
fix TOC errors, minor doco updates

### LINKED ISSUE #496: TOC links don't work in API docs
It's inconvenient to navigate through the API functions documentation because clicking on the table of contents links does not work due to missing underscores in the names of the corresponding destination anchors.

For example, https://github.com/nmoinvaz/minizip/blob/master/doc/mz_zip_rw.md#mz_zip_reader_open should be reachable through its TOC entry for `mz_zip_reader_open`, but it isn't, because the anchor name is spelled as `#mzzipreaderopen` (without underscores) in the TOC.

joycepg:
Fixes #496


nmoinvaz:
Thanks!

PR #502, opened 2020-06-24
Update zstd magic number to comply with the latest zip specification

fcharlie:
See: https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT

```txt
        0 - The file is stored (no compression)
        1 - The file is Shrunk
        2 - The file is Reduced with compression factor 1
        3 - The file is Reduced with compression factor 2
        4 - The file is Reduced with compression factor 3
        5 - The file is Reduced with compression factor 4
        6 - The file is Imploded
        7 - Reserved for Tokenizing compression algorithm
        8 - The file is Deflated
        9 - Enhanced Deflating using Deflate64(tm)
       10 - PKWARE Data Compression Library Imploding (old IBM TERSE)
       11 - Reserved by PKWARE
       12 - File is compressed using BZIP2 algorithm
       13 - Reserved by PKWARE
       14 - LZMA
       15 - Reserved by PKWARE
       16 - IBM z/OS CMPSC Compression
       17 - Reserved by PKWARE
       18 - File is compressed using IBM TERSE (new)
       19 - IBM LZ77 z Architecture 
       20 - deprecated (use method 93 for zstd)
       93 - Zstandard (zstd) Compression 
       94 - MP3 Compression 
       95 - XZ Compression 
       96 - JPEG variant
       97 - WavPack compressed data
       98 - PPMd version I, Rev 1
       99 - AE-x encryption marker (see APPENDIX E)

```

### DIFF
--- a/mz.h
+++ b/mz.h
@@ -63,8 +63,7 @@
 #define MZ_COMPRESS_METHOD_DEFLATE      (8)
 #define MZ_COMPRESS_METHOD_BZIP2        (12)
 #define MZ_COMPRESS_METHOD_LZMA         (14)
-#define MZ_COMPRESS_METHOD_ZSTD         (20)
-#define MZ_COMPRESS_METHOD_WZZSTD       (93)
+#define MZ_COMPRESS_METHOD_ZSTD         (93)
 #define MZ_COMPRESS_METHOD_AES          (99)
 
 #define MZ_COMPRESS_LEVEL_DEFAULT       (-1)
--- a/mz_zip.c
+++ b/mz_zip.c
@@ -1621,7 +1621,6 @@ static int32_t mz_zip_entry_open_int(void *handle, uint8_t raw, int16_t compress
 #endif
 #ifdef HAVE_ZSTD
     case MZ_COMPRESS_METHOD_ZSTD:
-    case MZ_COMPRESS_METHOD_WZZSTD:
 #endif
         err = MZ_OK;
         break;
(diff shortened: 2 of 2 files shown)


mz_strm_zstd.c
```diff
@@ -160,7 +160,7 @@ int32_t mz_stream_zstd_read(void *stream, void *buf, int32_t size) {
         zstd->total_in += in_bytes;
         zstd->total_out += out_bytes;
 
-    } while (zstd->in.pos < zstd->in.size && zstd->out.pos < zstd->out.size);
```
nmoinvaz: Why this change?

nmoinvaz:
We can just get rid of 0x20 value.

mz_strm_zstd.c
fcharlie: This is a previous modification and is currently consistent with upstream.

nmoinvaz:
Thanks!

gvollant:
I think this is a very bad idea to not support 20 magic when decompress

I'm of course OK to create new zipfile with magic 93.
But compatibility is always better than error.
So I prefer support both 20 and 93 for decompress zstd

gvollant:
@nmoinvaz @fcharlie

fcharlie:
@gvollant The key to this question is how many compression software/libraries support decompressing zip (zstd-20), but at present, I have not found any software or libraries that have been released to support magic number 20, so regarding compatibility, We don't need to consider the compatibility of magic number 20. Of course, if there is software that widely uses 20 as a zip(zstd) magic number, you can tell me. Thanks

fcharlie:
@gvollant Another problem is that the specifications in the zip format are in the hands of PKWARE, and interested developers cannot participate in the formulation of the specifications. The changes may be arbitrary.

nmoinvaz:
It looks like they changed the spec very quickly so it is doubtful that anybody used 20. Maybe a mistake on PKWARE’s part.

PR #505, opened 2020-06-24
Do not link Apple's Security framework when no crypt feature

### LINKED ISSUE #504: Don't link security library when crypto feature is disabled on Apple
if `MZ_PKCRYPT` and `MZ_WZAES` is OFF.
it's no need to link `CoreFoundation` and `Security` for Apple
there is condition of `(MZ_PKCRYPT OR MZ_WZAES)` for [WIN32](https://github.com/nmoinvaz/minizip/blob/master/CMakeLists.txt#L697) but no condition for Apple

okhowang:
fix #504


nmoinvaz:
I should have merged into `dev`. I have corrected it.

PR #513, opened 2020-07-16
Create cmake package config in 1.2 branch

mabuchner:
Create CMake package config + version file.

I tried not to departure too much from the code in the master branch.

The installation directory is different in the master branch, because the 1.2 branch already installs a `minizip-exports.cmake` file into a different directory.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -109,9 +109,39 @@ install(TARGETS minizip EXPORT minizip-exports
   LIBRARY DESTINATION "lib"
   ARCHIVE DESTINATION "lib")
 
+set(INSTALL_CMAKE_DIR "cmake")
 install(EXPORT minizip-exports
-        DESTINATION "cmake"
+        DESTINATION "${INSTALL_CMAKE_DIR}"
         NAMESPACE "MINIZIP::")
+
+# Create and install CMake package config version file to allow find_package()
+
+include(CMakePackageConfigHelpers)
+write_basic_package_version_file(
+  "${CMAKE_CURRENT_BINARY_DIR}/minizip-config-version.cmake"
+  VERSION 1.2
+  COMPATIBILITY SameMajorVersion)
+
+# Create config for find_package()
+
+file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/minizip-config.cmake.in"
+"
+@PACKAGE_INIT@
+include(CMakeFindDependencyMacro)
+find_dependency(ZLIB)
+include(\"\${CMAKE_CURRENT_LIST_DIR}/minizip-exports.cmake\")
+check_required_components(minizip)
+")
+
+configure_package_config_file(
+  "${CMAKE_CURRENT_BINARY_DIR}/minizip-config.cmake.in"
+  minizip-config.cmake
+  INSTALL_DESTINATION "${INSTALL_CMAKE_DIR}"
+  NO_SET_AND_CHECK_MACRO)
+
+install(FILES "${CMAKE_CURRENT_BINARY_DIR}/minizip-config-version.cmake"
+              "${CMAKE_CURRENT_BINARY_DIR}/minizip-config.cmake"
+        DESTINATION "${INSTALL_CMAKE_DIR}")
 
 install(FILES ${MINIZIP_PUBLIC_HEADERS}
   DESTINATION "include/minizip")


CMakeLists.txt
```diff
@@ -109,11 +109,36 @@ install(TARGETS minizip EXPORT minizip-exports
   LIBRARY DESTINATION "lib"
   ARCHIVE DESTINATION "lib")
 
+set(INSTALL_CMAKE_DIR "cmake")
```
mabuchner, line 112: The master branch installs the CMake config into `${CMAKE_INSTALL_FULL_LIBDIR}/cmake/${PROJECT_NAME}`, but the 1.2 branch installs the `minizip-exports.cmake` file into the `cmake` directory.

CMakeLists.txt
```diff
@@ -109,11 +109,36 @@ install(TARGETS minizip EXPORT minizip-exports
...
 install(EXPORT minizip-exports
-        DESTINATION "cmake"
+        DESTINATION "${INSTALL_CMAKE_DIR}"
         NAMESPACE "MINIZIP::")
 
-install(FILES ${MINIZIP_PUBLIC_HEADERS}
+# Create and install CMake package config version file to allow find_package()
+include(CMakePackageConfigHelpers)
+write_basic_package_version_file(${CMAKE_CURRENT_BINARY_DIR}/minizipConfigVersion.cmake
```
mabuchner, line 119: `minizip-config-version.cmake` would be more consistent with `minizip-exports.cmake`, but the master branch uses `minizipConfigVersion.cmake`.

nmoinvaz: If there is no expected way to name things, I don't mind changing master branch.

CMakeLists.txt
nmoinvaz: What about changing the names to `minizip-exports-1.2.cmake`? Or something along those lines? Or into a directory called `1.2`?

CMakeLists.txt
nmoinvaz: I have changed it in master `dev` branch.

CMakeLists.txt
```diff
@@ -109,10 +109,40 @@ install(TARGETS minizip EXPORT minizip-exports
...
+  "${CMAKE_CURRENT_BINARY_DIR}/minizipConfigVersion.cmake"
+  VERSION 1.2
+  COMPATIBILITY SameMajorVersion)
+
+# Create config for find_package()
+
+file(WRITE "${CMAKE_CURRENT_BINARY_DIR}/minizipConfig.cmake.in"
+"
+@PACKAGE_INIT@
+include(CMakeFindDependencyMacro)
```
nmoinvaz, line 130: What is this? And does the master branch need this chunk of cmake?

CMakeLists.txt
mabuchner: That’s kind of redundant information, because we already have a version config file.

My concern is not to be able to install multiple versions into the same installation directory. My concern is regarding consistency.

For consistency with the master branch we should install into ‘lib/cmake/minizip’, but someone might rely on the old installation directory.

If you think, this is not a problem, then I will change it.

CMakeLists.txt
mabuchner: Without the `find_dependency` function the consuming CMake project has to call `find_package(ZLIB)` in order to use minizip

```cmake
find_package(ZLIB)
find_package(minizip)
target_link_libraries(myapp PRIVATE MINIZIP::minizip)
```

With the `find_dependency` function the consuming project just needs to find the minizip package.

```cmake
find_package(minizip)
target_link_libraries(myapp PRIVATE MINIZIP::minizip)
```

That way someone using just minizip doesn't need to know about the fact, that it's internally using ZLIB.

Additionally, since ZLIB is a mandatory dependency for minizip the `find_dependency` function also ensures, that minizip is reported as "not found", if ZLIB wasn't found.

I think, it would be good, if the master branch also contains this chunk of cmake.

CMakeLists.txt
```diff
@@ -109,10 +109,40 @@ install(TARGETS minizip EXPORT minizip-exports
...
+find_dependency(ZLIB REQUIRED)
```
mabuchner, line 131: One thing, I just found out is, that `find_dependency` should not be called with `REQUIRED`, because this is supposed to get forwarded from the consuming code.

I will change this later.

CMakeLists.txt
```diff
@@ -109,11 +109,36 @@ install(TARGETS minizip EXPORT minizip-exports
...
+@PACKAGE_INIT@
+include(CMakeFindDependencyMacro)
+find_dependency(zlib REQUIRED)
+include(\"\${CMAKE_CURRENT_LIST_DIR}/minizip-exports.cmake\")
+")    
+
+# Create config for find_package()
+configure_package_config_file(
+  ${CMAKE_CURRENT_BINARY_DIR}/minizipConfig.cmake.in
+  minizipConfig.cmake
```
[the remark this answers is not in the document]
mabuchner: Renamed to `minizip-config.cmake`

CMakeLists.txt
mabuchner: Thanks. Renamed to `minizip-config-version.cmake`.

CMakeLists.txt
mabuchner: Removed `REQUIRED`.

mabuchner:
> If you update the filenames the way I did `dev` branch it would be good.

Renamed config files.

CMakeLists.txt
nmoinvaz: Ok I understand.

nmoinvaz:
Thank you!

PR #517, opened 2020-08-04
Unnecessary type conversions.

wzhengsen:


nmoinvaz:
This is a good change. Thanks for waiting.

PR #518, opened 2020-08-17
Guard the usage of __has_include behind the defined check

ItsBasi:
According to the gcc [doc](https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005finclude.html), compilers who do not support __hash_include evaluate the makro anyway, ignoring the is defined check.

Tried to compile the conan [recipe](https://github.com/conan-io/conan-center-index/pull/2415) with this profile

[settings]
arch=x86_64
arch_build=x86_64
build_type=Debug
compiler=Visual Studio
compiler.runtime=MTd
compiler.version=14
os=Windows
os_build=Windows
[options]
minizip:shared=False

### DIFF
--- a/mz.h
+++ b/mz.h
@@ -158,9 +158,13 @@
 #include <string.h> /* memset, strncpy, strlen */
 #include <limits.h>
 
-#if defined(HAVE_STDINT_H) || \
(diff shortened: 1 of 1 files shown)


nmoinvaz:
This is already fixed in `dev` branch.

PR #528, opened 2020-10-27
Allow option to skip git clone for MZ_BRG

UncleAlbie:
Downloading additional sources during build is problematic for
distribution integration.


nmoinvaz:
Have you thought about using `-DMZ_BRG=OFF -DMZ_OPENSSL=ON`? Does your platform use a different crypto library?

nmoinvaz:
This should be fixed with `-DMZ_FETCH_LIBS=OFF` option.

PR #529, opened 2020-10-31
mz_compat: sync parameter types with minizip1, add 2 compatibility wrappers

vszakats:
- zipOpenNewFileInZip*(): method/compression_method is 'int' in minizip1
- zipCloseFileInZipRaw64(): uncompressed_size is 'ZPOS64_T', which is 'uint64_t'
- unzTell64(): return value is 'ZPOS64_T', which is 'uint64_t'
- unztell(): return value should be 32-bit
- add unztell64() (full lowercase), which is the same as unzTell64()
- add unzeof(), which is the same as unzEndOfFile()

NOTE: It'd be probably useful to update mz_zip_entry_close_raw() uncompressed_size type to unsigned.

### DIFF
--- a/mz_compat.c
+++ b/mz_compat.c
@@ -139,7 +139,7 @@ void* zipGetStream_MZ(zipFile file) {
 
 int zipOpenNewFileInZip5(zipFile file, const char *filename, const zip_fileinfo *zipfi,
     const void *extrafield_local, uint16_t size_extrafield_local, const void *extrafield_global,
-    uint16_t size_extrafield_global, const char *comment, uint16_t compression_method, int level,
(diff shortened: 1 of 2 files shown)


nmoinvaz:
> NOTE: It'd be probably useful to update mz_zip_entry_close_raw() uncompressed_size type to unsigned.

Can you please explain?

vszakats:
This way the lower level type would match the type of higher level (compatibility) APIs, and it would also match the unsigned type used throughout the original minizip library. Given the file sizes involved, it's rather an academic than practical issue.

Thank you for merging!

nmoinvaz:
I think I am okay with leaving it as is for now. I use `int64_t` for compressed size and uncompressed size in a few places. Thanks a lot for your MR and work!

PR #530, opened 2020-11-18
Support building for midipix and LibreSSL 2.7.0+

Redfoxymoon:
midipix is a POSIX layer like cygwin for windows.

LibreSSL 2.7.0 and above has added the HMAC_CTX_{new,free,reset} functions.

### DIFF
--- a/mz_crypt_openssl.c
+++ b/mz_crypt_openssl.c
@@ -293,7 +293,7 @@ typedef struct mz_crypt_hmac_s {
 
 /***************************************************************************/
 
-#if (OPENSSL_VERSION_NUMBER < 0x10100000L) || defined(LIBRESSL_VERSION_NUMBER)
+#if (OPENSSL_VERSION_NUMBER < 0x10100000L) || (defined(LIBRESSL_VERSION_NUMBER) && (LIBRESSL_VERSION_NUMBER < 0x2070000fL))
 static HMAC_CTX *HMAC_CTX_new(void) {
     HMAC_CTX *ctx = OPENSSL_malloc(sizeof(HMAC_CTX));
     if (ctx != NULL)
--- a/mz_os.h
+++ b/mz_os.h
@@ -21,10 +21,10 @@ extern "C" {
 #  define MZ_VERSION_MADEBY_HOST_SYSTEM (MZ_HOST_SYSTEM_OSX_DARWIN)
 #elif defined(__riscos__)
 #  define MZ_VERSION_MADEBY_HOST_SYSTEM (MZ_HOST_SYSTEM_RISCOS)
-#elif defined(__unix__)
-#  define MZ_VERSION_MADEBY_HOST_SYSTEM (MZ_HOST_SYSTEM_UNIX)
 #elif defined(_WIN32)
 #  define MZ_VERSION_MADEBY_HOST_SYSTEM (MZ_HOST_SYSTEM_WINDOWS_NTFS)
+#else
+#  define MZ_VERSION_MADEBY_HOST_SYSTEM (MZ_HOST_SYSTEM_UNIX)
 #endif
 
 #if defined(HAVE_LZMA) || defined(HAVE_LIBCOMP)
--- a/mz_os_posix.c
+++ b/mz_os_posix.c
@@ -21,7 +21,7 @@
 #include <sys/types.h>
 #include <sys/stat.h>
 
-#if defined(__APPLE__) || defined(__unix__) || defined(__riscos__)
+#ifndef _WIN32
 #  include <utime.h>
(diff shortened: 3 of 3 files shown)


nmoinvaz:
It looks like not all CI have succeeded.

mz_os_posix.c
```diff
@@ -21,7 +21,7 @@
 #include <sys/types.h>
 #include <sys/stat.h>
 
-#if defined(__APPLE__) || defined(__unix__) || defined(__riscos__)
+#if defined(__APPLE__) || defined(__unix__) || defined(__midipix__) || defined(__riscos__)
```
nmoinvaz, line 24: This should be `#ifndef _WIN32` (or `_WINDOWS`, whatever I have used elsewhere).

mz_os.h
```diff
@@ -21,7 +21,7 @@ extern "C" {
 #  define MZ_VERSION_MADEBY_HOST_SYSTEM (MZ_HOST_SYSTEM_OSX_DARWIN)
 #elif defined(__riscos__)
 #  define MZ_VERSION_MADEBY_HOST_SYSTEM (MZ_HOST_SYSTEM_RISCOS)
-#elif defined(__unix__)
+#elif defined(__unix__) || defined(__midipix__)
```
nmoinvaz, line 24: This should be moved below `_WIN32`  and only use `#else`.

mz_crypt_openssl.c
```diff
@@ -293,7 +293,7 @@ typedef struct mz_crypt_hmac_s {
 
 /***************************************************************************/
 
-#if (OPENSSL_VERSION_NUMBER < 0x10100000L) || defined(LIBRESSL_VERSION_NUMBER)
+#if (OPENSSL_VERSION_NUMBER < 0x10100000L) || (LIBRESSL_VERSION_NUMBER < 0x2070000fL)
```
nmoinvaz, line 296: I think you still need to get check for existence of `LIBRESSL_VERSION_NUMBER`

Redfoxymoon: Fair enough, fixed.

mz_os_posix.c
Redfoxymoon: I am not sure I follow? mz_os_win32.c has system functions for win32, and mz_os_posix.c is not included when building for win32. Midipix is not a win32 target.

mz_os.h
Redfoxymoon: Fair enough, fixed:-)

mz_os_posix.c
nmoinvaz: max_os_posix.c is unique in that it should work on all platforms. Windows has a posix layer. By specifying `#ifndef _WIN32` it will include those files on all non-windows platforms which is essentially what is happening with `#if defined(__APPLE__) || defined(__unix__) || defined(__midipix__) || defined(__riscos__)`. So in order to prevent every non-Windows OS from having to add their own unique `defined` here we can just exclude windows.

mz_crypt_openssl.c
```diff
@@ -293,7 +293,7 @@ typedef struct mz_crypt_hmac_s {
 
 /***************************************************************************/
 
-#if (OPENSSL_VERSION_NUMBER < 0x10100000L) || defined(LIBRESSL_VERSION_NUMBER)
+#if (OPENSSL_VERSION_NUMBER < 0x10100000L) || defined(LIBRESSL_VERSION_NUMBER) && (LIBRESSL_VERSION_NUMBER < 0x2070000fL)
```
nmoinvaz, line 296: I think this is missing `()` around `defined(LIBRESSL_VERSION_NUMBER) && (LIBRESSL_VERSION_NUMBER < 0x2070000fL)`

mz_os.h
```diff
@@ -25,6 +25,8 @@ extern "C" {
 #  define MZ_VERSION_MADEBY_HOST_SYSTEM (MZ_HOST_SYSTEM_UNIX)
 #elif defined(_WIN32)
 #  define MZ_VERSION_MADEBY_HOST_SYSTEM (MZ_HOST_SYSTEM_WINDOWS_NTFS)
+#else
```
nmoinvaz, line 28: No need for `#elif defined(__unix__) above. We will assume all non-mentioned/non-detectible OSes as UNIX with our new `#else`.

Redfoxymoon: True.

mz_os_posix.c
Redfoxymoon: Alright, always nice to avoid unnecessary amounts of #ifdefs.

nmoinvaz:
It has been merged into `dev` for the next release. Thank you!

PR #531, opened 2020-11-24
mz_zip_rw.c - fix potential null dereference in create functions

eiszapfen2000:
In contrast to other mz_*_create() functions, mz_zip_reader_create and mz_zip_writer_create crash on a null argument. This fixes it.


nmoinvaz:
Thanks!

PR #532, opened 2020-11-25
CMakeLists.txt - fix zconf.h not being found in case of out-of-source…

eiszapfen2000:
…-tree build

zlib CMakeLists.txt does some strange things with zconf.h.in and zconf.h
To make sure zconf.h is found with an out-of-source-tree build, we have to adjust the add_subdirectory call, and give the source directory as binary directory.

ZLIB_BINARY_DIR is empty anyways, but that is caused by another issue in the clone_repo cmake helper, and even if it would have a valid value, for zlib we need to point to the source dir.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -177,7 +177,7 @@ elseif(MZ_ZLIB)
         clone_repo(ZLIB https://github.com/madler/zlib)
 
         # Don't automatically add all targets to the solution
-        add_subdirectory(${ZLIB_SOURCE_DIR} ${ZLIB_BINARY_DIR} EXCLUDE_FROM_ALL)
+        add_subdirectory(${ZLIB_SOURCE_DIR} ${ZLIB_SOURCE_DIR} EXCLUDE_FROM_ALL)
 
         list(APPEND MINIZIP_INC ${ZLIB_SOURCE_DIR})
         # Have to add zlib to install targets


ptc-tgamper:
Could we merge this. Until now we were keeping this one-liner around as a local patch, because we don't run CMake in mini zip's source directory.

nmoinvaz:
According to this line in the CMakeLists.txt for madler/zlib, it is configured to the binary directory.

https://github.com/madler/zlib/blob/cacf7f1d4e3d44d871b605da3b647f07d718623f/CMakeLists.txt#L84

What are the cmake commands you are using that reproduce the error?

eiszapfen2000:
I created a separate build directory and executed the follwing command in git bash:
`cmake ../minizip -DMZ_BZIP2=OFF -DMZ_LZMA=OFF -DMZ_ZSTD=OFF`

Afterwards I open the generated Visual Studio solution, and when building minzip I get the follwowing compile error:
`"zconf.h": No such file or directory`

From what I found, the following line in zlib cmake is related to the issue:
https://github.com/madler/zlib/blob/cacf7f1d4e3d44d871b605da3b647f07d718623f/CMakeLists.txt#L69

nmoinvaz:
It looks like FetchContent is using BINARY_DIR of the SOURCE_DIR.

CMakeLists.txt
```diff
@@ -177,7 +177,7 @@ elseif(MZ_ZLIB)
         clone_repo(ZLIB https://github.com/madler/zlib)
 
         # Don't automatically add all targets to the solution
-        add_subdirectory(${ZLIB_SOURCE_DIR} ${ZLIB_BINARY_DIR} EXCLUDE_FROM_ALL)
+        add_subdirectory(${ZLIB_SOURCE_DIR} ${ZLIB_SOURCE_DIR} EXCLUDE_FROM_ALL)
```
nmoinvaz, line 180: No need to specify `${ZLIB_SOURCE_DIR}` twice.

eiszapfen2000: Do you mean `add_subdirectory(${ZLIB_SOURCE_DIR} EXCLUDE_FROM_ALL)` would do the job?

nmoinvaz: Yes, if you change it I think we can merge it.

eiszapfen2000: Just tested it. It does not work with `add_subdirectory(${ZLIB_SOURCE_DIR} EXCLUDE_FROM_ALL)`, it really needs to be `add_subdirectory(${ZLIB_SOURCE_DIR} ${ZLIB_SOURCE_DIR} EXCLUDE_FROM_ALL)`

nmoinvaz: Thanks!

nmoinvaz:
This PR broke the Windows builds.

eiszapfen2000:
Could you provide the cmake line to reproduce?

nmoinvaz:
Actually it looks like for whatever reason cmake is finding an installed version of LZMA on Windows machines (probably GitHub changes the images) and it is failing. So not this PR.

CMakeLists.txt
nmoinvaz: > . The FetchContent_Populate call would set it as zlib_BINARY_DIR ...

Are you sure this is correct? The `name` argument passed to `clone_repo` is always uppercase.

eiszapfen2000:
Yep, looks like it:
https://github.com/nmoinvaz/minizip/runs/1507335552?check_suite_focus=true#step:7:51

CMakeLists.txt
eiszapfen2000: From the cmake documentation at https://cmake.org/cmake/help/latest/module/FetchContent.html

FetchContent_Populate() will set three variables in the scope of the caller; <lcName>_POPULATED, <lcName>_SOURCE_DIR and <lcName>_BINARY_DIR, where <lcName> is the lowercased <name>.

nmoinvaz: Actually docs for cmake shows it forces all lowercase..

> FetchContent_Populate() will set three variables in the scope of the caller; <lcName>_POPULATED, <lcName>_SOURCE_DIR and <lcName>_BINARY_DIR, where <lcName> is the lowercased <name>..

eiszapfen2000: Yeah, initially I changed all the code in the clone_repo to adjust for this, but in the end i still needed the one line fix in this PR. So I went with the minimal fix.

nmoinvaz:
I have made some additional changes that should fix things. Also I have added a CI build that checks that out-of-source-tree builds work.

eiszapfen2000:
Thank you, much appreciated

PR #533, opened 2020-12-01
CMakeLists.txt - Introduce MZ_ICONV

ptc-tgamper:
This adds an additional CMake option, that allows the user to disable iconv. We support non-Windows systems that have iconv, but we don't want to use it on that system, so adding this switch has been proven to be quite useful.


nmoinvaz:
I think this option also needs to be added to the README.md.

PR #551, opened 2021-02-01
fix: cmake find_package not working if lib is install in custom location

prateek9623:


CMakeLists.txt
```diff
@@ -722,8 +732,16 @@ if(NOT SKIP_INSTALL_LIBRARIES AND NOT SKIP_INSTALL_ALL)
     include(CMakePackageConfigHelpers)
     write_basic_package_version_file(${CMAKE_CURRENT_BINARY_DIR}/minizip-config-version.cmake
         COMPATIBILITY SameMajorVersion)
-
-    file(WRITE minizip-config.cmake.in "@PACKAGE_INIT@")
+    set(MINIZIP_CONFIG_CONTENT "@PACKAGE_INIT@")
+    if(MINIZIP_DEP_PKG)
+        string(APPEND MINIZIP_CONFIG_CONTENT "\ninclude(CMakeFindDependencyMacro)")
+        foreach(PKG_NAME ${MINIZIP_DEP_PKG})
+            message("asdasdasd ${PKG_NAME}")
```
nmoinvaz, line 739: I think some debugging info was left in.

CMakeLists.txt
```diff
@@ -722,8 +732,16 @@ if(NOT SKIP_INSTALL_LIBRARIES AND NOT SKIP_INSTALL_ALL)
     include(CMakePackageConfigHelpers)
     write_basic_package_version_file(${CMAKE_CURRENT_BINARY_DIR}/minizip-config-version.cmake
         COMPATIBILITY SameMajorVersion)
-
-    file(WRITE minizip-config.cmake.in "@PACKAGE_INIT@")
+    set(MINIZIP_CONFIG_CONTENT "@PACKAGE_INIT@")
+    if(MINIZIP_DEP_PKG)
+        string(APPEND MINIZIP_CONFIG_CONTENT "\ninclude(CMakeFindDependencyMacro)")
```
nmoinvaz, line 737: Can you place the null-terminators `\n` at the end of the lines instead of at the beginning?

CMakeLists.txt
prateek9623: oh, sorry about that

CMakeLists.txt
prateek9623: done

prateek9623:
I have also added alias target, using which 1 can include the target in almost all case whether it is included using find_package, external project/fetch content, or as git submodule

nmoinvaz:
I don't understand. Why do we need an `ALIAS` library? I think there is something I am missing. What is the need? Thanks.

prateek9623:
well if you look into generated "minizip.cmake" file, you will see there is no target with the name "minizip", there is one with "MINIZIP::minizip(line no 48 in minizip.cmake)".
image
due to: line no 731 in main CMakeLists.txt
image
So if you link target using find_package, you will only have the option to link with "MINIZIP::minizip". 
And also "target_link_libraries(main PUBLIC minizip)" only work when project included as subdirectory. So for consistency between find_package and subdirectory i added ALIAS target. 
Note:"target_link_libraries(main PUBLIC minizip)" works now also.

prateek9623:
So, now if one wants to link, they can do by:

- target_link_libraries(main PUBLIC minizip) -> only works if project is included as sub_directory
- target_link_libraries(main PUBLIC MINIZIP::minizip)-> works in almost all cases I think. (earlier was not working in subdirectory case)

CMakeLists.txt
```diff
@@ -678,6 +688,8 @@ set(INSTALL_PKGCONFIG_DIR "${CMAKE_INSTALL_FULL_LIBDIR}/pkgconfig"
 
 add_library(${PROJECT_NAME} ${MINIZIP_SRC} ${MINIZIP_HDR})
 
+add_library(MINIZIP::${PROJECT_NAME} ALIAS ${PROJECT_NAME})
```
nmoinvaz, line 691: Can you wrap this in: `if(NOT TARGET MINIZIP::${PROJECT_NAME})` in case user defines their own.

Then also call:
```
 set_target_properties(MINIZIP::${PROJECT_NAME} PROPERTIES
        INTERFACE_INCLUDE_DIRECTORIES "${CMAKE_CURRENT_SOURCE_DIR}");
```
That way if anybody links against `MINIZIP::${PROJECT_NAME}` they don't have to specify include dirs because it is already done. What do you think?

CMakeLists.txt
```diff
@@ -231,6 +232,7 @@ if(MZ_ZLIB)
     endif()
 
     if(MZ_ZLIB)
+        list(APPEND MINIZIP_DEP_PKG ZLIB)
```
nmoinvaz, line 235: Is it possible to use `MINIZIP_DEP` instead? It seems very similar?

nmoinvaz (commented):
These changes are looking good. Some more modifications. Sorry to be a pain.

CMakeLists.txt
nmoinvaz: Should this statement be moved to inside `if(ZLIB_FOUND AND NOT MZ_FORCE_FETCH_LIBS)` ? It seems like we only want to define dependent packages if we are using them. In the case of `MZ_FETCH_LIBS` the library is internally compiled.. 

Same with the rest?

prateek9623: > Is it possible to use `MINIZIP_DEP` instead? It seems very similar?

MINIZIP_DEP_PKG is the list of package names for  "find_package/find_dependency" and I think MINIZIP_DEP only gives target names, not the package name.
For e.g.:
ZLIB::ZLIB is the target, not package name, its package name is "ZLIB".  

> Should this statement be moved to inside `if(ZLIB_FOUND AND NOT MZ_FORCE_FETCH_LIBS)` ? It seems like we only want to define dependent packages if we are using them. In the case of `MZ_FETCH_LIBS` the library is internally compiled..

As you might know "find_dependency" checks for the dependency of minizip. So, even if the dependency is internally built we would still require it to find it because its config file is different from minizip.

nmoinvaz: I don't have much experience with `find_dependency`.

nmoinvaz:
Is this PR ready? It is still a draft.

prateek9623:
its ready

CMakeLists.txt
prateek9623: it is just as a check for dependencies before including package

nmoinvaz:
Thank you!

PR #554, opened 2021-02-11
fix_zstd_libraries_cmake

prateek9623:
Fixes zstd linking with find_package


CMakeLists.txt
```diff
@@ -343,21 +344,31 @@ if(MZ_ZSTD)
...
+                set(PC_PRIVATE_LIBS "${PC_PRIVATE_LIBS} -lzstd")
+                list(APPEND MINIZIP_DEP_PKG ZSTD)
+            endif()
         endif()
         if(NOT ZSTD_FOUND)
             find_package(ZSTD QUIET)
-            set(ZSTD_VERSION ${ZSTD_VERSION_STRING})
+            if(ZSTD_FOUND)
+                set(ZSTD_VERSION ${ZSTD_VERSION_STRING})
+                if(TARGET libzstd_static)
```
nmoinvaz, line 359: Perhaps this should be:
```
if(NOT DEFINED BUILD_SHARED_LIBS OR NOT ${BUILD_SHARED_LIBS})
```
Like on line 379.

nmoinvaz, line 359: Why not do: 
```
set(ZSTD_LIBRARIES libzstd_static)
```
Then we can use:
```
list(APPEND MINIZIP_LIB ${ZSTD_LIBRARIES})
```
down below in `if(ZSTD_FOUND AND NOT MZ_FORCE_FETCH_LIBS)` like before?

prateek9623: because there is not always the case that we would get shared libs, for e.g. if vcpkg is used to link zlib in Linux, then build would fail, because vcpkg don't support shared in linux

prateek9623: i can simplify this a bit, i just added this because I thought it would be pointless to add 
```cmake
                    set(PC_PRIVATE_LIBS "${PC_PRIVATE_LIBS} -lzstd")
                    list(APPEND MINIZIP_DEP_PKG ZSTD)
```
if we are linking statically, but after further thinking, it would fail if minizip is itself is static

nmoinvaz:
Thanks!

PR #555, opened 2021-02-22
Fix zstd libraries cmake

prateek9623:
was still giving an error in the previous pull request, this fixes that


nmoinvaz:
Looks good, can you rebase?

prateek9623:
@nmoinvaz i have a question when I include minizip with CMake's FetchContent it creates an error saying can't find libzstd_static/libzstd_shared location, can we anything about this. I suspect these lines

https://github.com/zlib-ng/minizip-ng/blob/6f17da582d2327c8cdc43622ddd31a185b6ef5fb/CMakeLists.txt#L364-L368

works inside Fetchcontent for building minizip but for parent project it can't find this libs

prateek9623:
> Looks good, can you rebase?

done I think

nmoinvaz:
Maybe try `list(APPEND MINIZIP_DEP $<TARGET_FILE:libzstd_shared>)`?

nmoinvaz:
Actually it looks like you are using `zstd::libzstd_shared` now. Is everything good for this PR?

prateek9623:
> Maybe try `list(APPEND MINIZIP_DEP $<TARGET_FILE:libzstd_shared>)`?

The problem is whenever minizip fetched zstd and minizip is included in another project using fetch content. Parent Project is not able to find  libzstd_shared/libzstd_static
These commits don't fix the above problem.

nmoinvaz:
Thank you!

PR #556, opened 2021-02-24
reflects-suffix-in-config-file

prateek9623:
adds suffix in config files to seperate installs from original minizip, and also helps in creating port file in vcpkg,


nmoinvaz:
Thanks!

PR #562, opened 2021-04-03
Fix inclusion of fetched bzip2 sources

jannkoeker:
FetchContent allows users to define the path to a prepopulated source folder by setting `FETCHCONTENT_SOURCE_DIR_<ID>`.
This will skip downloading and populating the repo, but also overrides `SOURCE_DIR`.
Hence you shouldn't make assumptions on where the fetched sources will be.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -262,24 +262,24 @@ if(MZ_BZIP2)
         # BZip2 repository does not support cmake so we have to create
         # the bzip2 library ourselves
         set(BZIP2_SRC
-            lib/bzip2/blocksort.c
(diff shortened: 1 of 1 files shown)


nmoinvaz:
Does this also need to be the same for the other third-party libraries?

jannkoeker:
Well, yes, but that is already the case for all the other third-party libraries. As they use the correct variable for `add_subdirectory`. Only the bzip2 sources were included without taking a changed source directory into account. So no further changes should be necessary.

nmoinvaz:
Nice job!

Issue #475, opened 2020-05-08
Windows: DLL export mz functions in mz_zip.h


jimpark:
On Linux and macOS, I can link dynamically to minizip and still access these symbols because they are exported by default. On Windows, they are not exported. Only the compatibility layer is exported. I'd like to move to start using the `mz` functions on all three platforms. Can we decorate the mz functions with MZ_EXPORT?

nmoinvaz:
It should be possible to call cmake with `-DCMAKE_WINDOWS_EXPORT_ALL_SYMBOLS=ON`.

jimpark:
Yes, it's just not consistent with the other header files that do have the export macro. If that's the solution, I'd suggest adding it directly to the CMakeLists.txt. It has no effect in linux or macOS but makes the shared library work on Windows. (One thing to note is that this CMake option does not export variables -- though I don't think we need to worry about it in minizip.)

zedxxx:
+1 for `MZ_EXPORT`

Currently dll compiled with mingw exports all symbols by default and there is a lot of "garbage" (functions from zlib etc). Adding export directive should fix this problem and we will have exported only `mz_` functions.

Here is a current `dumpbin` output:

```
Microsoft (R) COFF/PE Dumper Version 14.16.27026.1
Copyright (C) Microsoft Corporation.  All rights reserved.


Dump of file libminizip.dll

File Type: DLL

  Section contains the following exports for libminizip.dll

    00000000 characteristics
    5ED04684 time date stamp Fri May 29 02:17:24 2020
        0.00 version
           1 ordinal base
         463 number of functions
         463 number of names

    ordinal hint RVA      name

          1    0 00029820 BZ2_blockSort
          2    1 00021600 BZ2_bsInitWrite
          3    2 00021010 BZ2_bzBuffToBuffCompress
          4    3 00021130 BZ2_bzBuffToBuffDecompress
          5    4 0001EE50 BZ2_bzCompress
          6    5 0001EFC0 BZ2_bzCompressEnd
          7    6 0001EC00 BZ2_bzCompressInit
          8    7 0001F190 BZ2_bzDecompress
          9    8 00020180 BZ2_bzDecompressEnd
         10    9 0001F050 BZ2_bzDecompressInit
         11    A 00020CB0 BZ2_bzRead
         12    B 00020C10 BZ2_bzReadClose
         13    C 00020F80 BZ2_bzReadGetUnused
         14    D 00020810 BZ2_bzReadOpen
[...]

PR #565, opened 2021-04-13
Replace _WINDOWS by _WIN32

uilianries:
_WIN32 is commonly used and more supported than _WINDOWS pre definition.

The official predefines supported by Visual Studio are listed here: https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-160

This PR is related #475, but won't fix that, because only when MZ_COMPAT is enabled that minizip-ng provides external symbols.

/cc @impark @SpaceIm

### DIFF
--- a/mz.h
+++ b/mz.h
@@ -147,7 +147,7 @@
 #define MZ_FREE(PTR)                    (free(PTR))
 #endif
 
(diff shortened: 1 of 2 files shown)


nmoinvaz:
Thank you!

PR #574, opened 2021-05-17
Link zlib-ng native if available

Diatrus:


nmoinvaz:
I think one problem is that we don't know if system zlib-ng was built with `ZLIB_COMPAT` or not.

nmoinvaz:
Ah, actually if it has `-ng` suffix it was built with `ZLIB_COMPAT=OFF`

jhudsonWA:
Is my understanding correct that if we don't build with ZLIB_COMPAT we will have to include zlib-ng now? (if I don't I get a error that zlib-ng.h is missing)

PR #609, opened 2022-02-15
Fix buffer overrun in mz_path_combine

peterh:
The path may be overrun by two bytes. One because mz_path_append_slash
may append to path without updating path_len, and another one because
strncat always appends a NUL byte after writing N bytes.

Rescan the path length and subtract one from the strncat length parameter
to make sure we stay within max_path.

### DIFF
--- a/mz_os.c
+++ b/mz_os.c
@@ -33,7 +33,9 @@ int32_t mz_path_combine(char *path, const char *join, int32_t max_path) {
         path[max_path - 1] = 0;
     } else {
         mz_path_append_slash(path, max_path, MZ_PATH_SLASH_PLATFORM);
-        strncat(path, join, max_path - path_len);
+        path_len = (int32_t)strlen(path);
+        if (max_path > path_len)
+            strncat(path, join, max_path - path_len - 1);
     }
 
     return MZ_OK;
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,31 +11,31 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - name: Ubuntu 16 GCC 4.8
-            os: ubuntu-16.04
+          - name: Ubuntu 18 GCC 4.8
+            os: ubuntu-18.04
             compiler: gcc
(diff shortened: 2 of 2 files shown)


mz_os.c
```diff
@@ -33,7 +33,9 @@ int32_t mz_path_combine(char *path, const char *join, int32_t max_path) {
         path[max_path - 1] = 0;
     } else {
         mz_path_append_slash(path, max_path, MZ_PATH_SLASH_PLATFORM);
-        strncat(path, join, max_path - path_len);
+        path_len = (int32_t)strlen(path);
```
nmoinvaz, line 36: Can't we just do `patch_len += 1;` here?

peterh: We could do that. In the case where mz_path_append_slash does not append a slash (because it's already present), it means we will truncate the filename one character shorter than we need to.

I can respin that way if you think performance is more important than filling max_path.

nmoinvaz: I see. It's probably good for now the way it is.

nmoinvaz:
This doesn't seem to be passing the CI.

peterh:
Whoops. Reversed max_path test.

mz_os.c
```diff
@@ -33,7 +33,9 @@ int32_t mz_path_combine(char *path, const char *join, int32_t max_path) {
...
+        if (max_path > path_len)
```
nmoinvaz, line 37: Wouldn't we want to copy as many characters as possible? Here it only copies if it has enough room for all of it..

peterh: max_path is the capacity of `path`. path_len is what's already in `path`. So if path_len >= max_path, then path is already full and there is no room for any of it.

I would have omitted this check, except strncat takes a size_t (unsigned), so I worry that if the path is already full then `max_path - path_len - 1` will underflow due to the `- 1`, and strncat will try to copy infinite bytes.

mz_path_append_slash keeps the trailing NUL byte in place. So if we can assume that the caller will never pass in a max_path that is shorter than `path` already is, then we can omit this if statement. Let me know if you want me to make that assumption and remove the if().

PR #611, opened 2022-02-17
Fix unlikely buffer overruns

peterh:
One string length calculation in int16, and one missing NUL terminator.


nmoinvaz:
Thanks!

PR #615, opened 2022-03-09
[WIP] Improve requirements for zlib/zlib-ng dependencies if in-project zlib/zlib-ng targets exist

Demonese:
What problem does this solve:
```cmake
...

add_subdirectory(external/zlib) # zlib-ng
add_subdirectory(external/minizip)  # minizip-ng

...

add_executable(sample)
...
target_link_libraries(sample PRIVATE
    zlibstatic
    MINIZIP::minizip
)

...
```
In this case, minizip can automatically use the zlib-ng already in the project.


nmoinvaz:
What we do is:
```
add_subdirectory(zlib...)

# By setting these variables, `find_package` will use them to determine where zlib is.
set(ZLIB_ROOT ${CMAKE_CURRENT_SOURCE_DIR}/../lib/zlib CACHE PATH "Root path to local zlib")
set(ZLIB_LIBRARY $<TARGET_FILE:zlib> CACHE STRING "Link local zlib library")

add_subdirectory(minizip...)
```

Demonese:
> By setting these variables, `find_package` will use them to determine where zlib is.

Thanks, this method does work. I'm also trying to make it more automated.

Issue #636, opened 2022-07-07
slash behavior on Windows


jeremybernstein:
I may be doing something wrong, or overlooking a configuration option, but it appears that, on Windows systems, `MZ_PATH_SLASH_PLATFORM` is always going to be set to '\\'. This presents problems for archives created on Windows, but extracted on macOS or Unix, where '/' is expected (not to be an intolerable pedant, but the [.zip format description](https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT) specifies '/' in section 4.4.17:

```
       4.4.17.1 The name of the file, with optional relative path.
       The path stored MUST NOT contain a drive or
       device letter, or a leading slash.  All slashes
       MUST be forward slashes '/' as opposed to
       backwards slashes '\' for compatibility with Amiga
       and UNIX file systems etc.  If input came from standard
       input, there is no file name field.  
```

Mostly this makes `mz_zip_writer_add_path()` unusable on Windows (since it uses `mz_path_combine()` which then uses the platform slash), so it's something which can be worked around, but having an option to override this behavior would be welcome.

Thanks for considering it.

nmoinvaz:
Maybe a change here:
https://github.com/zlib-ng/minizip-ng/blob/47b8449fec2c7a575a506ecbe6399ad0603b5992/mz_zip.c#L790

nmoinvaz:
Merged to `develop` in #644.

PR #637, opened 2022-07-11
zip: conform slashes

jeremybernstein:
conform slashes to MZ_PATH_SLASH_UNIX before writing to zip stream. See #636

### DIFF
--- a/mz_zip.c
+++ b/mz_zip.c
@@ -788,12 +788,17 @@ static int32_t mz_zip_entry_write_header(void *stream, uint8_t local, mz_zip_fil
     }
 
     if (err == MZ_OK) {
-        char conformed_filename[512];
-        strncpy(conformed_filename, filename, sizeof(conformed_filename) - 1);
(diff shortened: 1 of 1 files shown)


mz_zip.c
```diff
@@ -787,12 +788,16 @@ static int32_t mz_zip_entry_write_header(void *stream, uint8_t local, mz_zip_fil
     }
 
     if (err == MZ_OK) {
-        if (mz_stream_write(stream, filename, filename_length) != filename_length)
+        char conformed_filename[512];
```
nmoinvaz, line 791: It would be better to just use `MZ_ALLOC` here, and if it returns NULL then return `MZ_ERROR_MEM`.

jeremybernstein: Thanks @nmoinvaz, modifications checked in.

nmoinvaz:
I have decided to go a different route in #644 so we don't have to allocate any memory.

Issue #640, opened 2022-07-27
Handle leak in mz_os_open_dir in low memory situations


aallidina-cb:
Thanks for maintaining this project. 

We ran minizip-ng against Coverity and found a few small issues. 

https://github.com/zlib-ng/minizip-ng/blob/47b8449fec2c7a575a506ecbe6399ad0603b5992/mz_os_win32.c#L402

In mz_os_win32.c::mz_os_open_dir(const char *path), the handle from FindFirstFileW gets leaked in low memory situations

```
    handle = FindFirstFileW(path_wide, &find_data);
    mz_os_unicode_string_delete(&path_wide);

    if (handle == INVALID_HANDLE_VALUE)
       // AA -> Should call FindClose(handle) before returning
        return NULL;

    dir_int = (DIR_int *)MZ_ALLOC(sizeof(DIR_int));
    if (dir_int == NULL)
        return NULL;
```

nmoinvaz:
I think you mean FindClose on second if statement return. Please open PR.

aallidina-cb:
Right, that's the one. OK, let me get to it.

nmoinvaz:
Fixed in `develop` branch.

PR #643, opened 2022-08-03
Fix find handle leak in low memory situations

aallidina-cb:
As described in issue #640


aallidina-cb:
Wow, I love the automated build checks.

I don't see any reason my change would trigger the failure though --
CMake / Windows GCC (pull_request) Failing after 2m — Windows GCC

"D:/a/minizip-ng/minizip-ng/lib/zlib/test/example.c:8:10: fatal error: zlib.h: No such file or directory
    8 | #include "zlib.h""

PR #644, opened 2022-08-06
Fixed writing backslashes in zip on Windows.

nmoinvaz:
All slashes should be written as forward slashes according
to the zip app note section 4.4.17.1.

Co-authored-by: jeremybernstein@users.noreply.github.com


jeremybernstein:
So happy to see this addressed. Thanks!

PR #648, opened 2022-08-12
build with libbsd-overlay

neheb:
Simplifies the code. Upstream also recommends it.

Signed-off-by: Rosen Penev <rosenp@gmail.com>

https://gitlab.freedesktop.org/libbsd/libbsd/-/issues/11#note_1417771


nmoinvaz:
Is `libbsd-overlay` available in all the same platforms? Do we need a fallback if `libbsd-overlay` does not exist?

nmoinvaz:
On Ubuntu 18 only `libbsd-dev` found in package search..

neheb:
libbsd-dev includes overlay. It's also supported on Windows.

PR #651, opened 2022-10-02
Switch getrandom() and arc4random_buf() usage order

thesamesam:
We need to match the order of inclusions at the top of the file otherwise we might end up trying to use arc4random_buf() when available (because HAVE_ARC4RANODM_BUF is set) even though we hit HAVE_GETRANDOM first above and only included <sys/random.h> because of it.

Besides, if getrandom() is available, we should really prefer it anyway.

Fixes an implicit function declaration:
```
minizip-ng-3.0.6/mz_os_posix.c:124:5: error: implicit declaration of function 'arc4random_buf' [-Werror=implicit-function-declaration]
```

Signed-off-by: Sam James <sam@gentoo.org>

### DIFF
--- a/mz_os_posix.c
+++ b/mz_os_posix.c
@@ -117,7 +117,22 @@ void mz_os_utf8_string_delete(uint8_t **string) {
 
 /***************************************************************************/
 
-#if defined(HAVE_ARC4RANDOM_BUF)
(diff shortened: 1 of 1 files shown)


thesamesam:
Thanks!

Issue #654, opened 2022-10-17
Support system copy of gtest


thesamesam:
Please consider reworking d03ca72a2520d7ba55e8763d14aa0d70b960ba84 to allow using a system copy of gtest. Many package managers prohibit network access during the build.

While it may be possible to pre-download a copy and shove into the build dir, it's still a waste of resources and awkward to maintain.

Thanks!

nmoinvaz:
It supports that already. If you have a parent project and already include Gtest then it will detect it.

thesamesam:
I don't think that helps when gtest is just installed on the system though? At least it doesn't for me, and other CMake projects pick it up fine.

I don't think the target test will ever pass as we're not doing find_package (or anything else) for it.

nmoinvaz:
Good idea. I am open to PRs.

nmoinvaz:
Thanks for the fix.

thesamesam:
No worries!

PR #657, opened 2022-11-19
Search system for GTest before downloading. #654

thesamesam:
Distributions often do builds with no network access available for both security reasons and also to ensure reproducibility.

This change tells CMake to query the system for a copy of gtest, but if it's not available, it'll fall back to downloading via FetchContent.


nmoinvaz:
Thanks!

thesamesam:
Thank you!

PR #659, opened 2022-11-27
CMake: call `project()` as soon as possible after `cmake_minimum_required()`

SpaceIm:
Currently, `project()` is called far too late in CMakeLists, almost at the end of the file. It can lead to many issues when users rely on toochain files (in conan it leads to several issues related to CMake policies). CMake advices to call project() ASAP after cmake_minimum_required().

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,10 @@
 #***************************************************************************
 
 cmake_minimum_required(VERSION 3.13)
+project(minizip-ng VERSION 3.0.7 LANGUAGES C)
+
+# API version
+set(MZ_SOVERSION "3")
 
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}/cmake")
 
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -975,9 +959,9 @@ if(MZ_BUILD_FUZZ_TESTS)
         set_target_properties(${target} PROPERTIES LINKER_LANGUAGE CXX)
         target_compile_definitions(${target} PRIVATE ${STDLIB_DEF})
         target_include_directories(${target} PRIVATE ${CMAKE_CURRENT_SOURCE_DIR})
-        target_link_libraries(${target} ${PROJECT_NAME})
+        target_link_libraries(${target} ${MZ_TARGET})
 
-        add_dependencies(${target} ${PROJECT_NAME})
+        add_dependencies(${target} ${MZ_TARGET})
```
Nekto89, line 964: isn't dependency already guaranteed after previous call to target_link_libraries?

SpaceIm: Indeed it's useless, but I prefer to not remove it in this PR to reduce reviewing complexity.

nmoinvaz:
Too many things are changed and renamed in one commit. What specific problem was this trying to fix?

SpaceIm:
Current logic where many things are done before first project() call defeats conan toolchain, and is not recommended by CMake: https://cmake.org/cmake/help/latest/command/project.html#usage. Basically it can lead to strange behaviors.

I'm sorry to have to change so many lines, but to be able to move project() near cmake_minimum_required() I had to:

- give a unique name to the project
- therefore all your ${PROJECT_NAME} logic must be replaced by another variable to preserve previous logic where target name (and therefore lib name) depends on an option. This is why there are so many changes, most of these changes are just a dumb find/replace.

kloczek:
https://github.com/zlib-ng/minizip-ng/pull/695 fixes proper use GNUInstallDirs cmake module.

nmoinvaz:
I have incorporated these changes into `develop`. Thanks.

PR #660, opened 2022-11-27
CMake: fix ALIAS target name to be consistent with imported target

SpaceIm:
Revert https://github.com/zlib-ng/minizip-ng/commit/9e9d8203e57f51542a1e46ee1ea2903029c88229, it's not consistent with imported target, which is `MINIZIP::minizip` if MZ_COMPAT is ON else `MINIZIP::minizip-ng`.

I don't know if you want to set ALIAS target AND imported target as `MINIZIP::minizip` unconditionally (`MINIZIP::minizip-ng` is probably a better choice), but if that's the case, you should use `EXPORT_NAME` property (indeed as per current CMakeLists, logical target name depends on PROJECT_NAME which depends on MZ_COMPAT, and by default export target name is logical target name) and also avoid to give a conditional name to CMake config file.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -685,6 +685,7 @@ set(INSTALL_PKGCONFIG_DIR "${CMAKE_INSTALL_LIBDIR}/pkgconfig"
     CACHE PATH "Installation directory for pkgconfig (.pc) files")
 
 add_library(${PROJECT_NAME} ${MINIZIP_SRC} ${MINIZIP_HDR})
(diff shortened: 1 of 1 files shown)


nmoinvaz (approved):

nmoinvaz:
> I don't know if you want to set ALIAS target AND imported target as MINIZIP::minizip unconditionally

Yes for backwards compatibility.

nmoinvaz:
I have incorporated these changes. Thanks!

PR #667, opened 2023-01-06
Fix private linking of LibLZMA

p2004a:
The current `liblzma` makes linke look for `libliblzma` as the `lib` prefix is added automatically. This is breaking static linking of library when using pkg-config for resolving dependencies.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -305,7 +305,7 @@ if(MZ_LZMA)
         list(APPEND MINIZIP_LIB ${LIBLZMA_LIBRARIES})
         list(APPEND MINIZIP_LBD ${LIBLZMA_LIBRARY_DIRS})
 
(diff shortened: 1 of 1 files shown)


p2004a:
I've also now noticed, that now in the pc file, between version 3.0.7 and 3.0.8, ssl started to be linked using absolute paths instead of -lssl and -lcrypto:

3.0.7:
```
Libs.private: -lz -lliblzma -lssl -lcrypto
```

3.0.8
```
Libs.private: -lz -lliblzma -l/home/p2004a/Workspace/BAR/spring/spring-static-libs/lib/libssl.a -l/home/p2004a/Workspace/BAR/spring/spring-static-libs/lib/libcrypto.a -l-lpthread -ldl
```

+ as you see the pthread is linked incorrectly.

nmoinvaz (approved):

PR #668, opened 2023-01-13
Support build with Swift Package Manager

luncliff:
I found my team's legacy is using `minizip`.
Currently, we are considering replacing it with this project.

Plus, if possible, we would like to import it from Swift sources.

### Changes

* Create `Package.swift` so the project can support https://www.swift.org/package-manager/

```console
user@host $ open Package.swift # Xcode will understand the file
```

```
swift build --configuration debug --triple arm64-apple-macosx
swift build --configuration release --triple x86_64-apple-macosx
```

#### Concerns

* Use built-in libraries like `libz.tbd`, `Security.framework`, etc.

### To Do

* [ ] Test import from Swift project

### DIFF
--- a/.github/workflows/swift-pm.yaml
+++ b/.github/workflows/swift-pm.yaml
@@ -10,3 +10,5 @@ jobs:
         run: swift build --verbose --configuration debug
       - name: "Swift Build(Release)"
         run: swift build --configuration release
+      - name: "Show Dependencies"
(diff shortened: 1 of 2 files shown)


nmoinvaz:
Sorry, I do not support any package management files in the repository.

luncliff:
I see. Thanks for the review :)

PR #673, opened 2023-02-14
Fix odd length key buffer overrun in mz_crypt_hmac_init()

sergio-nsk:


mz_crypt_win32.c
```diff
@@ -578,7 +579,7 @@ int32_t mz_crypt_hmac_init(void *handle, const void *key, int32_t key_length) {
         key_blob_s->hdr.bVersion = CUR_BLOB_VERSION;
         key_blob_s->hdr.aiKeyAlg = CALG_RC2;
         key_blob_s->hdr.reserved = 0;
-        key_blob_s->key_length = key_length;
+        key_blob_s->key_length = pad_key_length;
```
nmoinvaz, line 582: Shouldn't this still be `key_length`?

sergio-nsk: No, it should not, it is a key bloc with a 0-padded key.

nmoinvaz: Ok.

PR #677, opened 2023-03-09
Create parent directory before extracting symlink

jonathan-conder-sm:
At least when using Info-ZIP 3.0, symlinks to both files and directories lack a trailing slash, so the second mz_path_remove_filename is unnecessary (and could break things if the parent directory doesn't precede the symlink in the file). If other implementations do add a trailing slash, the if statement should handle that.

### DIFF
--- a/mz_zip_rw.c
+++ b/mz_zip_rw.c
@@ -719,7 +719,8 @@ int32_t mz_zip_reader_entry_save_file(void *handle, const char *path) {
     }
 
     /* If symbolic link then properly construct destination path and link path */
-    if (mz_zip_entry_is_symlink(reader->zip_handle) == MZ_OK) {
+    if ((mz_zip_entry_is_symlink(reader->zip_handle) == MZ_OK) &&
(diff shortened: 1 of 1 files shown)


mz_zip_rw.c
```diff
@@ -720,8 +720,10 @@ int32_t mz_zip_reader_entry_save_file(void *handle, const char *path) {
 
     /* If symbolic link then properly construct destination path and link path */
     if (mz_zip_entry_is_symlink(reader->zip_handle) == MZ_OK) {
-        mz_path_remove_slash(pathwfs);
-        mz_path_remove_filename(directory);
+        if (pathwfs[strlen(pathwfs) - 1] == MZ_PATH_SLASH_UNIX) {
```
nmoinvaz, line 723: Perhaps we can use `mz_path_has_slash` here?

jonathan-conder-sm: done, I also folded it into the previous `if`

PR #695, opened 2023-04-09
add proper use GNUInstallDirs cmake module

kloczek:
currentlly GNUInstallDirs cmake module is included but it does not work becaue there is no defined project language.
This commit adds peoper projct() entry and replace all custom cmake variables by `CMAKE_INSTALL_<foo>DIR` as it is described on https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html

Without proper project() cmake shows varning
```
CMake Warning (dev) at /usr/share/cmake/Modules/GNUInstallDirs.cmake:243 (message):
  Unable to determine default CMAKE_INSTALL_LIBDIR directory because no
  target architecture is known.  Please enable at least one language before
  including GNUInstallDirs.
```

Please update line with VERSION on release new version 😄


CMakeLists.txt
```diff
@@ -715,16 +710,16 @@ add_library(MINIZIP::minizip ALIAS ${PROJECT_NAME})
...
-        $<INSTALL_INTERFACE:${INSTALL_INC_DIR}>)
+        $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
 
     install(TARGETS ${PROJECT_NAME} ${MINIZIP_DEP}
             EXPORT ${PROJECT_NAME}
-            INCLUDES DESTINATION "${INSTALL_INC_DIR}"
-            RUNTIME DESTINATION "${INSTALL_BIN_DIR}"
-            ARCHIVE DESTINATION "${INSTALL_LIB_DIR}"
-            LIBRARY DESTINATION "${INSTALL_LIB_DIR}")
+            INCLUDES DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/minizip-ng"
```
nmoinvaz, line 717: Includes DIR should be different if using compat mode?

CMakeLists.txt
```diff
@@ -6,6 +6,11 @@
 #   schmieder.matthias@gmail.com
 
 cmake_minimum_required(VERSION 3.13)
+project(minuzip-ng
```
nmoinvaz, line 9: GitHub diff does not show where original `project` call is removed? 
Also project name is mis-spelled.
And original `project` call looks like:
```cmake
project(minizip${MZ_PROJECT_SUFFIX} LANGUAGES C VERSION ${VERSION})

```

CMakeLists.txt
kloczek: OK I've added commit to insytall headers in${CMAKE_INSTALL_INCLUDEDIR}/minizip-ng.
It was necessary update as well .pc file.

CMakeLists.txt
kloczek: Fixed that typo (sorry).
I'm not sure is it will be working with variable in project name.
If you know that it will work just merge that PR and continue 😄

kloczek: With multiline project() I think that it will be a bit easier to update VERSION.
I've addded as well project descriotion which could make project() line a bit to long if it will be in single line.

CMakeLists.txt
```diff
@@ -6,7 +6,7 @@
 #   schmieder.matthias@gmail.com
 
 cmake_minimum_required(VERSION 3.13)
-project(minuzip-ng
+project(minizip-ng
```
nmoinvaz, line 9: This ignores `MZ_PROEJCT_SUFFIX`..

kloczek:
BTW as bzip2, lzma and zstd provides pkgconfig intrface I thing that whole librarues detection oild be rplaced by `pkg_search_module()` calls
https://cmake.org/cmake/help/latest/module/FindPkgConfig.html#command:pkg_search_module

CMakeLists.txt
nmoinvaz: > I'm not sure is it will be working with variable in project name.

Why not?

> With multiline project()

I don't mind multiline.

kloczek: I don't know how such modyfication test a bit .. if you know that just please continue adding project() with variable.

nmoinvaz: Why change `project()` all?

CMakeLists.txt
kloczek: Only now I found that projest() is actually declared below my changes. Usually it is declared on top of the cmake files.
A bit odd way of using that statement.
I'll be busy in next day or two so feel free to close that PR and implement necessary chanhes basing on my PR.

nmoinvaz: I don't use any linux OS as my primary operating system, so my knowledge of install directories is rather limited.

nmoinvaz:
I have incorporated these changes the best I can. Thanks.

nmoinvaz:
> BTW as bzip2, lzma and zstd provides pkgconfig intrface I thing that whole librarues detection coiud be replaced by pkg_search_module()

I'm not so sure, I already use `pkg_check_module` and my CMake allows pulling those sources from their repos for systems such as Windows which doesn't have a well established package management system.

PR #698, opened 2023-05-04
Support AAD in AES-GCM

sergio-nsk:


mz_crypt_apple.c
```diff
@@ -25,6 +26,7 @@ enum {
 
 CCCryptorStatus CCCryptorGCMReset(CCCryptorRef cryptorRef);
 CCCryptorStatus CCCryptorGCMAddIV(CCCryptorRef cryptorRef, const void *iv, size_t ivLen);
+CCCryptorStatus CCCryptorGCMaddAAD(CCCryptorRef cryptorRef, const void *aData, size_t aDataLen);
```
nmoinvaz, line 29: Interesting, looks like this wrong cased name. But macOS docs, show
image
But odd that they both show the same macOS and iPhone minimum versions. Perhaps it was an internal name?

nmoinvaz: Let's change to correct case since it is the same `AVAILABLE_STARTING` version for both functions.

nmoinvaz:
I have removed the calls to `AES_*` in a separate commit. Some changes were missing from this PR, and it should be in a separate commit anyways. So this PR will need rebase. Sorry

sergio-nsk:
> I have removed the calls to `AES_*` in a separate commit. Some changes were missing from this PR, and it should be in a separate commit anyways. So this PR will need rebase. Sorry

You forgot to add `!aes-ctx` in parameter checks in `*encrypt*` and `*decrypt*` functions.

PR #702, opened 2023-05-30
Make minizip buildable for vitasdk

scribam:
Using the vitasdk it is possible to build minizip for the Playstation Vita, but it does not support position independent code.


Nekto89:
@nmoinvaz 
Maybe it would be even better to give full control to user through CMAKE_POSITION_INDEPENDENT_CODE instead of explicit usage of POSITION_INDEPENDENT_CODE?

nmoinvaz:
Yes that is a good idea.

Issue #705, opened 2023-06-03
Returning 'char *' from a function with result type 'uint8_t *'


NickBrighten:
Returning 'char *' from a function with result type 'uint8_t *' (aka 'unsigned char *') converts between pointers to integer types with different sign in mz_os_posix.c

nmoinvaz:
Should be resolved now in `develop` thanks to @ptc-tgamper.

PR #709, opened 2023-06-07
mz_os_posix.c - fix warning caused by missing cast of return value of strdup

ptc-tgamper:
I have run into https://github.com/zlib-ng/minizip-ng/issues/705, too.
Here is a fix.

### DIFF
--- a/mz_os_posix.c
+++ b/mz_os_posix.c
@@ -90,7 +90,7 @@ uint8_t *mz_os_utf8_string_create(const char *string, int32_t encoding) {
 }
 #else
 uint8_t *mz_os_utf8_string_create(const char *string, int32_t encoding) {
-    return strdup(string);
+    return (uint8_t*)strdup(string);
 }
 #endif
 


ptc-tgamper:
It seems all functions that deal with UTF-8 strings use uint8_t*. I must admit I like that because it explicitly tells me that all bits of the byte are relevant for the character.
I can have a look at generally changing the UTF-8 strings to use char*, but that means changing the code of the Win32 implementation, too.

mz_os.h
```diff
@@ -107,10 +107,10 @@ wchar_t *mz_os_unicode_string_create(const char *string, int32_t encoding);
 void     mz_os_unicode_string_delete(wchar_t **string);
 /* Delete a unicode string that was created */
 
-uint8_t *mz_os_utf8_string_create(const char *string, int32_t encoding);
+char    *mz_os_utf8_string_create(const char *string, int32_t encoding);
 /* Create a utf8 string from a string with another encoding */
 
-void     mz_os_utf8_string_delete(uint8_t **string);
+void     mz_os_utf8_string_delete(char **string);
```
ptc-tgamper, line 113: API break

mz_os_win32.c
```diff
@@ -58,41 +58,41 @@ void mz_os_unicode_string_delete(wchar_t **string) {
...
-    uint8_t *string_utf8 = NULL;
+char *mz_os_utf8_string_create_from_unicode(const wchar_t *string, int32_t encoding) {
+    char *string_utf8 = NULL;
     uint32_t string_utf8_size = 0;
 
     MZ_UNUSED(encoding);
 
     string_utf8_size = WideCharToMultiByte(CP_UTF8, 0, string, -1, NULL, 0, NULL, NULL);
-    string_utf8 = (uint8_t *)calloc(string_utf8_size + 1, sizeof(wchar_t));
+    string_utf8 = (char *)calloc(string_utf8_size + 1, sizeof(char));
```
ptc-tgamper, line 87: Same here, sizeof(wchar_t) vs sizeof(char)

ptc-tgamper:
CI failure is caused by https://git.tukaani.org/xz.git/ not responding

nmoinvaz:
LGTM

Issue #669, opened 2023-01-31
Split archive bug


mileta-r:
When creating a split archive minizip on Windows10 machine creates thousands of empty files.
To reproduce unpack the attached file to a "SplitZipTest" folder and try to create a new archive with:
   minizip.exe -v -5 -k 200 SplitZip.zip path\to\SplitZipTest\\*
[SplitZip.zip](https://github.com/zlib-ng/minizip-ng/files/10551438/SplitZip.zip)

nmoinvaz:
This should be resolved now in the `develop` branch. Thanks.

PR #710, opened 2023-06-07
mz_zip.c - correctly close entry with streams other than split_stream

ptc-tgamper:
This fixes a bug caused by https://github.com/zlib-ng/minizip-ng/issues/669

Above introduced the only instance where we actually check the return value of `mz_stream_set_prop_int64()`. Given any stream that is not a split stream, the call to `mz_stream_set_prop_int64(MZ_STREAM_PROP_DISK_NUMBER, ..)` will fail. So, do the same as in other places, ignore the return value of `mz_stream_set_prop_int64()`, but still check everything else.


nmoinvaz:
LGTM

Issue #690, opened 2023-04-06
Names truncated at 256 bytes


pmqs:
While looking at #689 I noticed a number of arrays (`path` , `utf8_name` and `resolved_name` ) with hard-wired lengths in `mz_zip_reader_save_all`. The code that uses these arrays looks like it can silently truncate output filenames when the overall path being written exceeds these array lengths.

The attached zip file, [test.zip](https://github.com/zlib-ng/minizip-ng/files/11171416/test.zip), contains a patnh with two  150 character directory names. Individually they are ok, but the combined lengths will result is a path that blows the 256 byte length of the `utf8_name` array.

```
$ unzip -l test.zip 
Archive:  test.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
        5  2023-04-06 14:47   base/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx/yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy/text.txt
---------                     -------
        5                     1 file
```

The ultimate payload, `text.txt`,  contains this

```
$ unzip -p test.zip 
abcd
```

Lets see what `unzip` extracts 

```
$ unzip test.zip 
Archive:  test.zip
[...]

PR #716, opened 2023-06-11
Fix for Names truncated at 256 bytes

pmqs:
Partial fix for https://github.com/zlib-ng/minizip-ng/issues/690

Heavy use of `malloc`/`realloc` in this, so have tested this for leaks/corruption by hacking sanitizer into the build. There are some leaks around `mz_crypt_sha_create` and `mz_zip_writer_entry_open` that were already there, so haven't made it worse.

Can post the sanitizer patch if you like.


pmqs:
Created https://github.com/zlib-ng/minizip-ng/issues/717 to report the leaks and hacked sanitizer build is in https://github.com/pmqs/minizip-ng/tree/sanitizer

nmoinvaz:
I am second guessing whether or not just to use malloc or not. Because alloca requires headers depending on the platform. And also can cause problem/exeception if there isn't enough stack space. I agree it is unlikely, but who knows.  alloca is more efficent if using lots of small files in a zip though.

pmqs:
Yep, I was in two minds about that myself. In the end I decided to avoid hitting the heap when I could keep it all on the stack with  `alloca`.  It will be exceptional use-cases of very long filenames/paths (which this fix is designed to address) that will be the problem.

If you think `malloc` is likely to have less portability issues, I can redo the change easily enough.

pmqs:
Changed the `alloca` to `malloc` for now.

nmoinvaz:
I have cleaned up and incorporated these changes.

Issue #717, opened 2023-06-11
Sanitizer reports memory leaks in `mz_crypt_sha_create` and `mz_zip_writer_entry_open`


pmqs:
Hacked sanitizer build  to test fix for #690 shows existing leaks in `mz_crypt_sha_create` and `mz_zip_writer_entry_open`. Not sure if these are openssl issues or issues with this code.

see https://github.com/pmqs/minizip-ng/actions/runs/5236387911/jobs/9453896686

```
6/211 Test   #6: raw-erase-generic ................***Failed    0.38 sec
=================================================================
==2374==ERROR: LeakSanitizer: detected memory leaks
Direct leak of 624 byte(s) in 26 object(s) allocated from:
    #0 0x7fdc54b5ea37 in __interceptor_calloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:154
    #1 0x55fe86dd8e36 in mz_crypt_sha_create /home/runner/work/minizip-ng/minizip-ng/mz_crypt_openssl.c:251
    #2 0x55fe86dcea7a in mz_zip_writer_entry_open /home/runner/work/minizip-ng/minizip-ng/mz_zip_rw.c:1333
    #3 0x55fe86dd261c in mz_zip_writer_copy_from_reader /home/runner/work/minizip-ng/minizip-ng/mz_zip_rw.c:[17](https://github.com/pmqs/minizip-ng/actions/runs/5236387911/jobs/9453896686#step:8:18)52
    #4 0x55fe86db68e7 in minizip_erase /home/runner/work/minizip-ng/minizip-ng/minizip.c:477
    #5 0x55fe86db39a6 in main /home/runner/work/minizip-ng/minizip-ng/minizip.c:657
    #6 0x7fdc54299d8f  (/lib/x86_64-linux-gnu/libc.so.6+0x29d8f)
Indirect leak of 3640 byte(s) in 32 object(s) allocated from:
    #0 0x7fdc54b5e867 in __interceptor_malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:145
    #1 0x7fdc5464f41d in CRYPTO_zalloc (/lib/x86_64-linux-gnu/libcrypto.so.3+0x1b741d)
Indirect leak of [18](https://github.com/pmqs/minizip-ng/actions/runs/5236387911/jobs/9453896686#step:8:19)72 byte(s) in 26 object(s) allocated from:
[...]

PR #718, opened 2023-06-12
Fix for memory leak 

pmqs:
This fixes https://github.com/zlib-ng/minizip-ng/issues/717 and includes sanitizer support to allow verification (i just borrowed the cmake code from zlib-ng).

Don't really know my way around the code, so the fix I provided may not be the best/proper way to do it.


nmoinvaz:
I have merged this change in manually.

PR #725, opened 2023-08-01
Do not override error

NielsBillen:
Only override the error value when `err` is equal to `MZ_OK`. Otherwise, when providing the wrong password for an encrypted file, the error value will be `MZ_MEM_ERROR` instead of `MZ_PASSWORD_ERROR`


nmoinvaz:
It looks like this has already been resolved in the `develop` branch.

PR #726, opened 2023-08-21
Allow negative zstd compression level

ahupna:
It is not documented properly: https://github.com/facebook/zstd/issues/3133
But zstd compression level can go negative, and larger than int16_t as well: https://news.ycombinator.com/item?id=31428760 

Changed the preset type in the mz_stream_zstd to be a signed int32_t and removed < 0 clamping when setting level.

### DIFF
--- a/mz_strm_zstd.c
+++ b/mz_strm_zstd.c
@@ -326,6 +326,7 @@ void *mz_stream_zstd_create(void) {
     if (zstd) {
         zstd->stream.vtbl = &mz_stream_zstd_vtbl;
         zstd->max_total_out = -1;
+        zstd->preset = ZSTD_CLEVEL_DEFAULT;
     }
     return zstd;
 }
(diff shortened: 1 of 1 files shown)


nmoinvaz:
I am currently using `MZ_COMPRESS_LEVEL_DEFAULT (-1)`, so how does this work with that? Does that value need to be changed to something else?

mkasick:
I looked at this while working on https://github.com/zlib-ng/minizip-ng/pull/730, and I'm not sure the complexity of shoehorning support for negative levels as part of minizip's command-line options is worthwhile.  The use case for negative levels is to support compression of high-bandwidth streaming data, which isn't within scope of an archive file format.  I think support for zstd's "regular" levels of 1 through 22 is sufficient.

nmoinvaz:
I have made some commits that allow for negative values (just not -1), through the API. It is still not available through the command line. If somebody wants to tackle that, I am open to it, but having the double dashes for command line args is a bit weird when it is normally only single dash.

PR #729, opened 2023-09-05
Fix pkgconfig file

autoantwort:
If libz-ng is used the declared dependency on libz was wrong. Also it should be private if `Libs.private` is used. And if the libs are available as pkgconfig file we should use them instead of hardcoding liker flags.


nmoinvaz:
Thanks!

PR #730, opened 2023-09-12
Support compression levels >= 10 and use zstd's internal default level.

mkasick:
zstd supports regular compression levels 1 through 22.  This change enables minizip to accept command-line options of "-10" and above to set levels higher than 10.

This also changes minizip to send the special level of 0 when no command-line level option is specified, which indicates to use zstd's internal default level (currently 3, but potentially subject to change).

zstd also supports negative compression level values to enable "fast" compression.  These levels disable Huffman coding to achieve high bandwidth for streaming loads (comparable to LZ4).  I didn't attempt to support negative levels here as I'm not sure they're of much value in a non-streaming archive format.


nmoinvaz:
Thanks!

PR #743, opened 2023-11-17
Reject overflows of zip header fields in minizip compat

### LINKED ISSUE #736: Filenames or file comments >= 65536 bytes are truncated
When adding a file to a ZIP archive or file comment whose name is >= 65536 bytes, it is truncated. A name equal to 65536 bytes exactly is stored as having an empty name.

These condition should result in an error, not the creation of a corrupted ZIP file.

The attached program will show this issue by creating a file with a name 65536 bytes long.

[badzip.c.txt](https://github.com/zlib-ng/minizip-ng/files/13059978/badzip.c.txt)

tbeu:
This PR applies https://github.com/madler/zlib/pull/843 to minizip compat to close #736. It applies the [patch](https://github.com/madler/zlib/pull/843.patch) provided by @zmodem, but removed the checks for unused extra fields.

> This checks the lengths of the file name, and comment that would be put in the zip headers, and rejects them if they are too long. They are each limited to 65535 bytes in length by the zip format. This also avoids possible buffer overflows if the provided fields are too long.

### DIFF
--- a/mz_compat.c
+++ b/mz_compat.c
@@ -464,9 +464,9 @@ int zipOpenNewFileInZip5(zipFile file, const char *filename, const zip_fileinfo 
         return ZIP_PARAMERROR;
 
     // The filename and comment length must fit in 16 bits.
-    if ((filename != NULL) && (strlen(filename) > 0xffff))
+    if (filename && strlen(filename) > 0xffff)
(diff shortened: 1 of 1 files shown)


Coeur:
Landed in minizip-ng 4.0.4

PR #790, opened 2024-07-05
Do not use REQUIRED with find_dependency

dg0yt:
Noticed while updating the vcpkg port:

`find_dependency` handles forwarding of `REQUIRED` from the original `find_package` call.
An explicit `REQUIRED` breaks non-required `find_package`, and it stops the regular config mode search behaviour.


nmoinvaz:
Thank you for your contribution.

PR #804, opened 2024-10-21
fix: avoid private api on Mac App Store

Coeur:


nmoinvaz:
I'm not sure __APPLE__ makes sense since these files are for Apple platform.

nmoinvaz:
I think we need something like this which is by default `ON`.
```cmake
cmake_dependent_option(MZ_BUILD_FOR_APPSTORE "Build without macOS private APIs" ON "APPLE" OFF)
```

Coeur:
OK. Quick question @nmoinvaz, since you need those declarations for a private project: do you want GCM on all Apple platforms (including iOS)?
- if yes, I'll use `MZ_TARGET_APPSTORE` to replace TARGET_OS_IPHONE
- if no, I'll combine `MZ_TARGET_APPSTORE` with `!TARGET_OS_IPHONE` so that GCM only builds for macOS.

nmoinvaz:
> if no, I'll combine MZ_TARGET_APPSTORE with !TARGET_OS_IPHONE so that GCM only builds for macOS.

This one. 

Thank you!

Coeur:
Actually, I prefer to default it to 0 for all platforms, as it would be hard to see more users interested in a non-distributable version of minizip instead of a distributable version of minizip.

nmoinvaz (approved):

mz_crypt_apple.c
```diff
@@ -20,8 +20,12 @@
 
 /***************************************************************************/
 
-/* Avoid use of private API for iOS, Apple does not allow it on App Store. Zip format doesn't need GCM. */
-#if !TARGET_OS_IPHONE
+#ifndef MZ_TARGET_APPSTORE
+#define MZ_TARGET_APPSTORE 0
```
Coeur, line 24: I think I should have defaulted it to 1, @nmoinvaz .
Can you add a commit to flip the default value?

nmoinvaz: Done

PR #807, opened 2024-10-30
Group compatibilities per origin

Coeur:
As the documented purpose of zlib-ng is to be a drop-in replacement of zlib, and minizip-ng a drop-in replacement of minizip, I wanted to clarify which parts of mz_compat.h are for the original minizip compatibility (permanent compatibility), and which parts are for older minizip-ng compatibility (deprecated usage).

I've consulted the current headers at https://github.com/madler/zlib/tree/develop/contrib/minizip and I've grouped them accordingly.

### DIFF
--- a/mz_compat.h
+++ b/mz_compat.h
@@ -111,11 +111,14 @@ typedef struct zlib_filefunc64_def_s
 
 /***************************************************************************/
 
+/* Compatibility layer with the original minizip library (ioapi.h and iowin32.h). */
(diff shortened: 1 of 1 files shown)


Coeur:
I don't know what's happening with the Ubuntu builds.

nmoinvaz:
I was just thinking about this this morning. I think it might be better if we create a `compat` folder and split `mz_compat.h/c` into their separate files in that folder.

nmoinvaz:
I have fixed the disk splitting issue, if you want to rebase on top of `develop`.

nmoinvaz (approved):

PR #809, opened 2024-10-30
Split compat layer into different source files

nmoinvaz:


VVD:
This commit break backward compatibility in patch-level update. This is very a "bad practice".
Break build consumers, for example https://github.com/qmc2/qmc2-mame-fe:
```
src/imagewidget.h:13:10: fatal error: 'mz_compat.h' file not found
   13 | #include "mz_compat.h"
      |          ^~~~~~~~~~~~~
```
It have dozen places with `#include "mz_compat.h"`.
Details are here: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=284326

Can you please return back this file?

nmoinvaz:
If you are using compat mode I would think you would also be using headers from original minizip??

PR #812, opened 2024-11-01
Adopting a clang-format

Coeur:
This is my attempt to define rules which are matching the existing code style as close as possible.
The rule definitions are available at:
https://clang.llvm.org/docs/ClangFormatStyleOptions.html

And you can run such command to apply the corrections:
```
clang-format -i **/*.{c,h}
```

The result isn't perfect, hence why it's just a draft. Although I'm clueless about how to do better.

--

We could independently cherry-pick the "good" fixes, like:
- mz_os_posix.c lines 274-278
- mz_strm.c line 379
- mz_strm.h line 102
- mz_strm_buf.c line 57
- mz_strm_mem.c line 214
- mz_strm_pkcrypt.c line 165
- mz_zip.c line 2783

--

A viable alternative to clang-format would be Uncrustify which has more options:
https://github.com/uncrustify/uncrustify/blob/master/documentation/htdocs/default.cfg


nmoinvaz:
We use `clang-format` at work and start with the Google style.

What do you think of this:
```
---
BasedOnStyle: Google

AccessModifierOffset: -2
AlignConsecutiveMacros: AcrossEmptyLines
AlignOperands: DontAlign
AllowShortFunctionsOnASingleLine: None
AllowShortIfStatementsOnASingleLine: Never
AllowShortLoopsOnASingleLine: false
ColumnLimit: 120
DerivePointerAlignment: false
IndentCaseLabels: false
IndentWidth: 4
InsertNewlineAtEOF: true
IndentPPDirectives: AfterHash
ObjCBlockIndentWidth: 4
PointerAlignment: Right
PPIndentWidth: 2
SortIncludes: Never
```

Coeur:
I've taken your proposal, but I kept the consecutive alignments, which is already in use and more readable.

nmoinvaz:
I'm a fan of the alignment in structs but not alignment of variable declarations in functions and I don't think there is a way to do one and not the other with clang-format.

Coeur:
OK, well, fine for me.
I won't apply the clang-format changes myself: I'll let you do it, as you're the prime contributor of most affected lines of code.

nmoinvaz (approved):

PR #815, opened 2024-11-01
Expose zlib constants for backwards compat

### LINKED ISSUE #747: Backwards compatibility issue with minizip-ng
I have run into a project that has an embedded copy of minizip 1.x. It tries to use an external copy if found but fails to build with minizip 4.x.

It uses the zip.h / unzip.h headers which in minizip 4.x I see are empty and just use mz_compat.h.

The build trips up trying to find the constants Z_BEST_COMPRESSION and Z_DEFAULT_STRATEGY.

The minizip 1.x zip.h / unzip.h headers include the zlib.h header, the minizip 4.x headers do not.

I would expect that the minizip 4.x headers would provide the constants if they're part of the minizip API.

Coeur:
Fix #747
Supersedes #750 (cc @brad0)

I re-did the PR to account for the recent refactor in 7df56f7cc8438b1b67c881cad049b29bf075bccd

### DIFF
--- a/compat/ioapi.h
+++ b/compat/ioapi.h
@@ -5,8 +5,11 @@
 #include <stdint.h>
 
 #ifndef _ZLIB_H
-#  include "zlib.h"
-#endif
+#  if __has_include(<zlib.h>)
+#    include <zlib.h>
+#  elif __has_include(<zlib-ng.h>)
+#    include <zlib-ng.h>
+#  endif
 #endif
 
 typedef uint64_t ZPOS64_T;
--- a/compat/ioapi.h
+++ b/compat/ioapi.h
@@ -4,11 +4,11 @@
 
 #include <stdint.h>
 
-#ifndef _ZLIB_H
-#  if __has_include(<zlib.h>)
-#    include <zlib.h>
-#  elif __has_include(<zlib-ng.h>)
+#if !defined(_ZLIB_H) && !defined(ZLIB_H) && !defined(ZLIB_H_)
+#  if __has_include(<zlib-ng.h>)
 #    include <zlib-ng.h>
+#  elif __has_include(<zlib.h>)
+#    include <zlib.h>
 #  endif
 #endif
 
--- a/compat/ioapi.h
+++ b/compat/ioapi.h
@@ -7,12 +7,6 @@
 #ifndef _ZLIB_H
 #  include "zlib.h"
 #endif
-
-#ifndef Z_ERRNO
-#define Z_ERRNO                         (-1)
-#endif
-#ifndef Z_DEFLATED
-#define Z_DEFLATED                      (8)
 #endif
 
 typedef uint64_t ZPOS64_T;
--- a/compat/ioapi.h
+++ b/compat/ioapi.h
@@ -5,7 +5,14 @@
 #include <stdint.h>
 
 #ifndef _ZLIB_H
-//#include "zlib.h"
+#  include "zlib.h"
+#endif
+
+#ifndef Z_ERRNO
+#define Z_ERRNO                         (-1)
+#endif
+#ifndef Z_DEFLATED
+#define Z_DEFLATED                      (8)
 #endif
 
 typedef uint64_t ZPOS64_T;
--- a/mz_crypt.c
+++ b/mz_crypt.c
@@ -34,11 +34,15 @@ int32_t mz_crypt_rand(uint8_t *buf, int32_t size) {
 
 uint32_t mz_crypt_crc32_update(uint32_t value, const uint8_t *buf, int32_t size) {
 #if defined(HAVE_ZLIB)
-    /* Define z_crc_t in zlib 1.2.5 and less or if using zlib-ng */
-#  if (ZLIB_VERNUM < 0x1270)
-    typedef unsigned long z_crc_t;
-#  else
+#  ifndef ZLIB_VERNUM
+    /* HAVE_ZLIB but no ZLIB_VERNUM? */
     typedef uint32_t z_crc_t;
+#  elif (ZLIB_VERNUM & 0xf != 0xf) && (ZLIB_VERNUM < 0x1270)
+    /* Define z_crc_t in zlib 1.2.6 and less */
+    typedef unsigned long z_crc_t;
+#  elif (ZLIB_VERNUM & 0xf == 0xf) && (ZLIB_VERNUM < 0x12df)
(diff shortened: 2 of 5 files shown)


compat/ioapi.h
```diff
@@ -5,7 +5,14 @@
...
-//#include "zlib.h"
+#include "zlib.h"
+#endif
+
+#ifndef Z_ERRNO
+#define Z_ERRNO                         (-1)
+#endif
+#ifndef Z_DEFLATED
+#define Z_DEFLATED                      (8)
 #endif
```
Coeur, line 16: I can also remove this entirely.
After all, we don't attempt to redefine `Z_DEFAULT_STRATEGY` which we use and which is also coming from zlib.h

mz_crypt.c
```diff
@@ -37,8 +37,6 @@ uint32_t mz_crypt_crc32_update(uint32_t value, const uint8_t *buf, int32_t size)
    /* Define z_crc_t in zlib 1.2.5 and less or if using zlib-ng */
 #  if (ZLIB_VERNUM < 0x1270)
     typedef unsigned long z_crc_t;
-#  else
-    typedef uint32_t z_crc_t;
 #  endif
```
Coeur, line 40: z_crc_t was added in:
- zlib 1.2.7 (https://github.com/madler/zlib/commit/6c9bd474aa08312ef2e2e9655a80e18db24a1680) as `unsigned long`
- zlib-ng 2.1.0 (https://github.com/zlib-ng/zlib-ng/commit/ff9162c1a0da8d60ae664e5930b256de99ad83b3) as `uint32_t`

They were added with conflicting types, so we should not redefine them.

mz_crypt.c
```diff
@@ -34,11 +34,12 @@ int32_t mz_crypt_rand(uint8_t *buf, int32_t size) {
...
-#  if (ZLIB_VERNUM < 0x1270)
+#  if (ZLIB_VERNUM & 0xf != 0xf) && (ZLIB_VERNUM < 0x1270)
+    /* Define z_crc_t in zlib 1.2.6 and less */
     typedef unsigned long z_crc_t;
-#  else
-    typedef uint32_t z_crc_t;
+#  elif (ZLIB_VERNUM & 0xf == 0xf) && (ZLIB_VERNUM < 0x12df)
+    /* Define z_crc_t in zlib-ng 2.0.7 and less */
+    typedef unsigned int z_crc_t;
 #  endif
```
Coeur, line 43: z_crc_t was added in:
- zlib 1.2.7 (https://github.com/madler/zlib/commit/6c9bd474aa08312ef2e2e9655a80e18db24a1680) as `unsigned long`
- zlib-ng 2.1.0 (https://github.com/zlib-ng/zlib-ng/commit/ff9162c1a0da8d60ae664e5930b256de99ad83b3) as `unsigned int`

They were added with conflicting types, so we should not redefine them.

CMakeLists.txt
```diff
@@ -701,7 +701,9 @@ target_include_directories(${MINIZIP_TARGET} PUBLIC
 if(MZ_COMPAT)
     target_include_directories(${MINIZIP_TARGET} PUBLIC
         $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/compat/>
-        $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>)
+        $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>
+        ${ZLIB_SOURCE_DIR}
+        ${ZLIB_BINARY_DIR})
```
Coeur, line 706: This seems to cause build issues for Windows.
I don't know how to fix it.

brad0:
Thanks for taking this on.

brad0:
Also I noticed anything even remotely modern for zlib uses ```ZLIB_H``` instead of ```_ZLIB_H``` for the header guard, which is kinda strange as such guards usually have a prefix underscore.

https://github.com/madler/zlib/commit/8e34b3a8024c028dd9fd21d70525fc6d215efde5

Coeur:
Oh my, good catch.
- the legacy zlib uses `ZLIB_H` since zlib 1.2.0.2. It was using `_ZLIB_H` before that. (https://github.com/madler/zlib/commit/8e34b3a8024c028dd9fd21d70525fc6d215efde5)
- zlib-ng uses `ZLIB_H_` since zlib-ng 1.9.9-b1. It was using `ZLIB_H` before that. (https://github.com/zlib-ng/zlib-ng/commit/d1de3909245b9152993cf8dced0c4d69fe04efd4)

brad0:
Also regarding header guards..

unzip.h / zip.h
_ZLIBIOAPI_H

ioapi.h
ZLIBIOAPI64_H

Coeur:
> Also regarding header guards..

Those aren't a big deal. And if we want to change it, it's not in the scope of this PR.

compat/ioapi.h
```diff
@@ -4,8 +4,12 @@
 
 #include <stdint.h>
 
-#ifndef _ZLIB_H
-//#include "zlib.h"
+#if !defined(_ZLIB_H) && !defined(ZLIB_H) && !defined(ZLIB_H_)
+#  if __has_include(<zlib-ng.h>)
```
nmoinvaz, line 8: I'm afraid `__has_include` is not available in MSVC?

nmoinvaz: Perhaps it will only work if included into C++ file?

Coeur: `__has_include` is in VS 2017 15.3: https://learn.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance?view=msvc-170

Do we care about obsolete versions of Visual Studio which are more than 7 years old?

AannhTonn:
.

compat/ioapi.h
AannhTonn: .

Coeur: @nmoinvaz can we merge this PR as it is, and deal with VS separately?
Otherwise, I guess, all what is needed is to add some nearly empty compat.cc file which just contains `include "compat.h"`? But I'm worried it could affect other platforms which don't expect C++ code.

nmoinvaz (approved):

nmoinvaz:
Since compat files moved to compat dir, not sure how to get cmake to use those when installing.

AannhTonn:
xin  chào tôi cần giúp  đỡ từ  bạn

AannhTonn:
.

PR #826, opened 2024-11-11
compatibility with Swift Package Manager

Coeur:
Swift Package Manager (which is supported by https://github.com/ZipArchive/ZipArchive and https://github.com/marmelroy/Zip), requires include paths to be matching the structure on disk.


nmoinvaz:
I think the compat dir needs to be added to include directories.

Coeur:
OK. Workaround with `.headerSearchPath("minizip")`.

PR #827, opened 2024-11-13
ignore number_entry_cd

### LINKED ISSUE #822: Archive cannot be unzipped with `unzOpen_MZ` but works with `mz_zip_reader_open_file`
[Books.zip](https://github.com/user-attachments/files/17701115/Books.zip)

This archive can be opened with `mz_zip_reader_open_file`, but cannot be opened with `unzOpen_MZ`.

I also tried adding `mz_zip_set_recover(handle, 1);` to `unzOpen_MZ`, but that wasn't enough to succeed at opening the file.

It can be opened with native macOS Archive Utility (aka, double-click to open the zip file on macOS).

Coeur:
Fix #822.

I opted for ignoring number_entry_cd instead of attempting to recover cd because the cd recovery method doesn't work when the split_stream isn't configured. `mz_zip_recover_cd` is going to use `mz_stream_os_tell` instead of `mz_stream_split_tell` and recovery will fail.

An alternative to my fix would be:
1. First to add `zip->recover = 1` inside `unzOpen_MZ`,
2. and second to add some split_stream support inside `unzOpen_MZ` ... which I don't know how to.

(if you know how to add split_stream support to `unzOpen_MZ`, then feel free to supersede my fix with yours.

Although, in my opinion, the streaming mode ignores CD, so why bothering verifying if `number_entry_cd` is valid?


nmoinvaz:
I will need some time to think about this.

Coeur:
On a note, the sample archive that we're trying to open was shared 6 years ago here: https://github.com/ZipArchive/ZipArchive/issues/478
And there is no context where it came from or what tool was used to make it.
@LiuDeng can you tell use where that Books.zip is from, or how it was built?

AannhTonn:
......

Coeur:
@nmoinvaz I've improved the logic to only skip number_entry_cd  when recover is true.
And I've added a minimal test file in the fuzz folder.

nmoinvaz (approved):

AannhTonn:
....?

AannhTonn:
có ai không liên lạc cho tôi với....

AannhTonn:
có ai không liên lạc lại cho tôi với....?

PR #833, opened 2024-11-23
fix macOS CI

Coeur:
Similar fix as the one I did for another project: https://github.com/transmission/transmission/pull/7254


nmoinvaz:
It looks like last week they deployed new runners, so this is not necessary anymore.
https://github.com/actions/runner-images/issues/10984#issuecomment-2501326144

PR #834, opened 2024-11-24
fix: unzipping with backslashes in the name on UNIX

Coeur:
While #832 was for fixing zipping, this PR is for fixing unzipping.

The expectation is that archiving+unarchiving data on UNIX should preserve its structure, no matter if there are backslashes (or dots) in the name of files or folders.
So backslash should only be seen as a separator on Windows, never on UNIX filesystems.

This is the current behaviour with the default macOS `/usr/bin/zip` and `/usr/bin/unzip` as well as default behaviour in Finder when right-click -> compress, followed by double-click to uncompress.


Coeur:
OK, I guess I need to look at the unit tests.

nmoinvaz:
The thing I worry about is if the zip was improperly created in the Windows style of path slashes in the zip entry. It would unzip on Unix anymore.

Coeur:
Do we have Windows sample archives like that? As written in 4.4.17.1, it would not have been allowed in the first place.

OK, here is what I suggest: a compile option just like I did in #804 with MZ_TARGET_APPSTORE. Flipping this compile option would allow for either behaviour.

Coeur:
OK, next mission, a compile option...

Coeur:
I've added MZ_PRESERVE_NATIVE_STRUCTURE in https://github.com/zlib-ng/minizip-ng/pull/834/commits/216e242b51d8b3b046ac0b14f771130cff1db111

Coeur:
> The thing I worry about is if the zip was improperly created in the Windows style of path slashes in the zip entry. It would unzip on Unix anymore.

If the filepath was previously `a/b\\c/d\\e`, the Windows style or archiving would have made it `a\\b\\c\\d\\e` so that archive would have been already corrupted and is irrecoverable to its original structure, no matter which compilation flag you use. Hopefully, with #832 we shouldn't have corrupted archives anymore.

nmoinvaz (approved):

Coeur:
For reference, this is the (expected) impact of this PR.

Without MZ_PRESERVE_NATIVE_STRUCTURE:

| Input              | Archive format  | Unzip on Linux              | Unzip on Windows      |
| ------------- | --------------- | ---------------------- | ---------------------- |
| Windows A\b | A/b                     | A/b                                 | A\b                                 |
| Linux A/b       | A/b                     | A/b                                 | A\b                                 |
| Linux a\\b      | a\\b                     | a/b [non-preserved]     | a\b [non-preserved]    |
| Linux A/b\\c   | A/b\\c                 | A/b/c [non-preserved] | A\b\c [non-preserved] |

With MZ_PRESERVE_NATIVE_STRUCTURE:

| Input              | Archive format  | Unzip on Linux  | Unzip on Windows      |
| ------------- | --------------- | --------------- | ---------------------- |
| Windows A\b | A/b                     | A/b                     | A\b                                |
| Linux A/b       | A/b                     | A/b                     | A\b                                |
| Linux a\\b      | a\\b                     | a\\b                    | a\b [non-preserved]    |
| Linux A/b\\c   | A/b\\c                 | A/b\\c                | A\b\c [non-preserved] |

mz_os.h
```diff
@@ -150,6 +150,9 @@ struct dirent *mz_os_read_dir(DIR *dir);
 int32_t mz_os_close_dir(DIR *dir);
 /* Closes a directory that has been opened for listing */
 
+int32_t mz_os_is_dir_separator(const char c);
```
sergio-nsk, line 153: It looks like a copy/paste unnecessary `const`.

Coeur: #911

PR #844, opened 2025-02-28
Align structures for 64-bit platforms

GermanAizek:
Fully similar issue as in zlib-ng: https://github.com/zlib-ng/zlib-ng/pull/1881

- mz_stream_split 120 -> 104 bytes
- mz_zip_file, mz_zip_entry 128 -> 120 bytes
- minizip_opt 32 -> 24 bytes
- mz_crypt_sha 32 -> 24 bytes
- mz_crypt_hmac 40 -> 32 bytes

### DIFF
--- a/minizip.c
+++ b/minizip.c
@@ -23,16 +23,16 @@
 /***************************************************************************/
 
 typedef struct minizip_opt_s {
+    int64_t disk_size;
+    int32_t encoding;
     uint8_t include_path;
     int16_t compress_level;
     uint8_t compress_method;
     uint8_t overwrite;
     uint8_t append;
-    int64_t disk_size;
     uint8_t follow_links;
     uint8_t store_links;
     uint8_t zip_cd;
-    int32_t encoding;
     uint8_t verbose;
     uint8_t aes;
 } minizip_opt;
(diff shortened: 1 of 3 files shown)


mz_zip.h
```diff
@@ -30,12 +30,12 @@ typedef struct mz_zip_file_s {
     time_t modified_date;        /* last modified date in unix time */
     time_t accessed_date;        /* last accessed date in unix time */
     time_t creation_date;        /* creation date in unix time */
-    uint32_t crc;                /* crc-32 */
```
nmoinvaz, line 33: I don't think this structure can be changed due to compatiblity.

nmoinvaz (changes requested):

mz_zip.h
GermanAizek: > I don't think this structure can be changed due to compatiblity.

@nmoinvaz, im fixed this in commit: https://github.com/zlib-ng/minizip-ng/pull/844/commits/83e0e5462ea2dc5dc0579b4632fa92300777bdfe

nmoinvaz (approved):

PR #845, opened 2025-02-28
Pre-calc length destination directory before while loop

GermanAizek:
Before Clang 19 -O2:

Pre-calc length destination directory before while loop

After Clang 19 -O2:

Pre-calc length destination directory before while loop

### DIFF
--- a/mz_zip_rw.c
+++ b/mz_zip_rw.c
@@ -825,6 +825,7 @@ int32_t mz_zip_reader_save_all(void *handle, const char *destination_dir) {
     if (err == MZ_END_OF_LIST)
         return err;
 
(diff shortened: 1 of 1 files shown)


mz_zip_rw.c
```diff
@@ -825,14 +825,15 @@ int32_t mz_zip_reader_save_all(void *handle, const char *destination_dir) {
     if (err == MZ_END_OF_LIST)
         return err;
 
+    int destination_dir_len = strlen(destination_dir) + 1;
```
nmoinvaz, line 828: int32_t and move declaration to top of function?

nmoinvaz (changes requested):

PR #849, opened 2025-03-15
Same CMake Build zlib

tanolino:
I wanted to build zlib-ng within the same project, but I don't want to make a build-install step for that. The fetch function also did not work. So here I propose to use an existing zlib target, found in the current cmake project


nmoinvaz:
You should be able to use `-D MZ_FORCE_FETCH_LIBS=OFF` then create a CMake alias for `ZLIB::ZLIB` that is picked up by `find_library(ZLIB)`.

The information on this page may be of some use to you:
https://github.com/zlib-ng/zlib-ng/wiki/Common-build-problems

PR #856, opened 2025-04-09
Make `pattern` in `mz_zip_reader` dynamically assigned

DreamHelium:
Since `pattern` should be used along with the usage of functions like `mz_zip_reader_goto_next_entry`, and one could put a dynamically-assigned `pattern` in another function, which makes it difficult to free the original string, there's need to make the `pattern` in `mz_zip_reader` dynamically assigned, and the old string can be freed immediately.


mz_zip_rw.c
```diff
@@ -230,6 +230,11 @@ int32_t mz_zip_reader_close(void *handle) {
         mz_stream_delete(&reader->mem_stream);
     }
 
+    if (reader->pattern) {
```
Coeur, line 233: On line 915 you didn't put any `if`, so why putting one here? You can probably remove it from line 233.

DreamHelium: I referred to the `mz_zip.c` code, which is the similar code in `mz_zip_set_comment` and `mz_zip_close`. But the `if` in line 233 seems unnecessary indeed, I could remove it later.

PR #857, opened 2025-04-13
gitgnore build folders

Coeur:
When I run those commands, I get a lot of folders created:
```
cmake -S . -B . -D MZ_BUILD_TESTS=ON -D MZ_BUILD_UNIT_TESTS=ON -D BUILD_SHARED_LIBS=OFF -D CMAKE_BUILD_TYPE=Debug
cmake --build . --config Debug
ctest --output-on-failure -C Debug
```

So I'm adding those folders to .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,10 @@
+# builds
+_deps/
 build/
+CMakeFiles/
+lib/
+Testing/
 third_party/
+
+# IDE
 .vscode


nmoinvaz (approved):

Coeur:
Windows build failure seems unrelated:
>fatal: unable to access 'https://sourceware.org/git/bzip2.git/': Failed to connect to sourceware.org port 443 after 21072 ms: Could not connect to server

Nekto89:
IMO in-source builds are usually not a good practice

Coeur:
> IMO in-source builds are usually not a good practice

In my opinion, having all the source and headers at the root of the project is also not a great practice. That clutters the front page of a project.

PR #862, opened 2025-04-16
Using MZ_COMPAT doesn't match old Minizip output

### LINKED ISSUE #861: Using MZ_COMPAT doesn't match old Minizip output
Platform: Windows (but compiling for Android with git-bash
Compiler: clang++ w/Ninja via CMake
Versions: minizip-ng 4.0.9 and zlib-ng 2.2.4

I've been updating our build system to use zlib-ng & minizip-ng and while I started by trying to use the non-compatible versions, I quickly realized I was going to be stuck with the backwards compatible APIs for a while until some dependencies update. So I switched and started directing all my dependency builds to point to the -ng pair.

I've run into a couple of strange problems though.

1. When using MZ_COMPAT, crypt.h isn't copied into the output folder from `compat`
2. The find_package helpers don't have the correct minizip include path; they include `minizip` in the include path when they should be pointing at the folder below that (e.g. `/install/location/include` vs `/install/location/include/minizip`)

(1) I think this just requires

```cmake
    list(APPEND MINIZIP_HDR
        compat/ioapi.h
        compat/unzip.h
        compat/zip.h)
```
to become
```cmake
    list(APPEND MINIZIP_HDR
        compat/crypt.h
        compat/ioapi.h
        compat/unzip.h
        compat/zip.h)
```
I don't see a downside to this change, though it's possible there's some setting somewhere that would cause that header to included and I've somehow missed it.

(2) I checked the output of the original minizip and it's package helper has this block in it:

```cmake
set_target_properties(MINIZIP::minizip PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
  ...
)
```
whereas the package helper from minizip-ng writes out this block:

```cmake
set_target_properties(MINIZIP::minizip PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include/minizip;${_IMPORT_PREFIX}/include/minizip"
  ...
)
```
I'm not sure why it repeats it twice, but if you change

```cmake
    install(TARGETS ${MINIZIP_TARGET} ${MINIZIP_DEP}
            EXPORT ${MINIZIP_TARGET}
            INCLUDES DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/${MINIZIP_TARGET}"
            RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}"
            ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}"
            LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}")
```
to
```cmake
    install(TARGETS ${MINIZIP_TARGET} ${MINIZIP_DEP}
            EXPORT ${MINIZIP_TARGET}
            INCLUDES DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}"
            RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}"
            ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}"
            LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}")
```
you get this improved output:
```cmake
set_target_properties(MINIZIP::minizip PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include/minizip;${_IMPORT_PREFIX}/include"
  ...
)
```
Still not sure where the first one is coming from, but this would seem to cover both cases if they were important I think.

I'm going to make these changes on my private copy of the repo but I think these changes are worth considering, unless of course I'm doing something wrong. I don't _think_ I am though, my build is pretty generic (variables point to local builds of dependencies):

```bash
   -DMZ_COMPAT=ON
   -DMZ_FETCH_LIBS=OFF
   -DMZ_OPENSSL=ON
   -DOPENSSL_ROOT_DIR=$openssl_root_dir
   -DOPENSSL_INCLUDE_DIR=$openssl_root_dir/include
   -DOPENSSL_CRYPTO_LIBRARY=$openssl_root_dir/lib/libcrypto.a
   -DOPENSSL_SSL_LIBRARY=$openssl_root_dir/lib/libssl.a
   -DCMAKE_FIND_ROOT_PATH=$zlib_root_dir
```

If you want a PR I can whip one up pretty easily, these are small changes but I thought I'd check first I wasn't missing something or that there was some ramification of these that I haven't noticed.

AndrewAtAvenza:
* Removed extra 'minizip' or 'minzip-ng' from inherited include path in cmake package
* Added missing 'crypt.h' (from 'compat/crypt.h') when installing after building with MZ_COMPAT

Resolves #861

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -622,6 +622,7 @@ if(MZ_COMPAT)
         compat/unzip.c
         compat/zip.c)
     list(APPEND MINIZIP_HDR 
+        compat/crypt.h
         compat/ioapi.h
         compat/unzip.h
         compat/zip.h)
@@ -711,7 +712,7 @@ if(NOT SKIP_INSTALL_LIBRARIES AND NOT SKIP_INSTALL_ALL)
 
     install(TARGETS ${MINIZIP_TARGET} ${MINIZIP_DEP}
             EXPORT ${MINIZIP_TARGET}
-            INCLUDES DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/${MINIZIP_TARGET}"
+            INCLUDES DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}"
             RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}"
             ARCHIVE DESTINATION "${CMAKE_INSTALL_LIBDIR}"
             LIBRARY DESTINATION "${CMAKE_INSTALL_LIBDIR}")


CMakeLists.txt
```diff
@@ -711,7 +712,7 @@ if(NOT SKIP_INSTALL_LIBRARIES AND NOT SKIP_INSTALL_ALL)
 
     install(TARGETS ${MINIZIP_TARGET} ${MINIZIP_DEP}
             EXPORT ${MINIZIP_TARGET}
-            INCLUDES DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}/${MINIZIP_TARGET}"
+            INCLUDES DESTINATION "${CMAKE_INSTALL_INCLUDEDIR}"
```
nmoinvaz, line 715: I wonder what the historial reason for ${MINIZIP_TARGET} is? Can you check? I'm not feeling well.

Nekto89: https://github.com/zlib-ng/minizip-ng/commit/e15e1462f2d6c9b9e9cd63b979223f4f835de40b

nmoinvaz: Perhaps we should remove this change here as it could affect Linux distributions?

nmoinvaz: Say if a target has both minizip and minizip-compat? It would then create two dirs??

nmoinvaz: Maybe we can add multiple INCLUDES DESTINATIONS?

CMakeLists.txt
```diff
@@ -672,11 +617,12 @@ endif()
...
     set(SOVERSION "1")
 
-    list(APPEND MINIZIP_SRC
+    list(APPEND MINIZIP_SRC 
         compat/ioapi.c
         compat/unzip.c
         compat/zip.c)
-    list(APPEND MINIZIP_HDR
+    list(APPEND MINIZIP_HDR 
+        compat/crypt.h
```
nmoinvaz, line 625: This one-line change is merged as part of #982. Thanks.

nmoinvaz:
The other changes in this PR I think might be too distruptive.

PR #867, opened 2025-05-01
fix: unzipping with backslash at the end of the filename

Coeur:
Testing, on UNIX system:
```
touch a\\
./minizip file.zip a\\
rm a\\
./minizip -x file.zip
ls a\\
```


nmoinvaz:
Hi @Coeur nice fix. I think the commit message needs more description for the history. Thanks!

nmoinvaz (approved):

PR #874, opened 2025-05-09
Feature: dynamic pattern for selected files

### LINKED ISSUE #855: Need to keep string's lifetime when using set-param functions
I was used to using the dynamic string in my project, so when calling functions like `mz_zip_reader_set_pattern`, I couldn't free the string after calling such functions, which **must** cause a segfault. Keeping the lifetime of such string in function's lifetime might be hard since one can forget to free such string.
Since freeing a NULL pointer is safe, it is possible to make strings in structure dynamically assigned and then freed at last.

Coeur:
Fix: #855

The difference with #856 is that we're moving the `free(reader->pattern)` from `mz_zip_reader_close` to `mz_zip_reader_delete`.

cc @DreamHelium

### DIFF
--- a/mz_zip_rw.h
+++ b/mz_zip_rw.h
@@ -102,7 +102,7 @@ int32_t mz_zip_reader_save_all(void *handle, const char *destination_dir);
 
 /***************************************************************************/
 
-void mz_zip_reader_set_pattern(void *handle, const char *pattern, uint8_t ignore_case);
+int32_t mz_zip_reader_set_pattern(void *handle, const char *pattern, uint8_t ignore_case);
 /* Sets the match pattern for entries in the zip file, if null all entries are matched */
 
 void mz_zip_reader_set_password(void *handle, const char *password);
--- a/doc/mz_zip_rw.md
+++ b/doc/mz_zip_rw.md
@@ -805,13 +805,13 @@ Sets the match pattern for entries in the zip file, if null all entries are matc
 **Return**
 |Type|Description|
 |-|-|
-|void|No return|
+|int32_t|[MZ_ERROR](mz_error.md) code, MZ_OK if successful|
 
 **Example**
 ```
 int32_t matches = 0;
 const char *pattern = "*.txt";
-mz_zip_reader_set_pattern(zip_reader, pattern, 1);
+if (mz_zip_reader_set_pattern(zip_reader, pattern, 1) == MZ_OK)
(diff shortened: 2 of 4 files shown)


nmoinvaz (approved):

Coeur:
You don't need to merge it immediately. I'll search for a better fix in the next 48h.

Coeur:
I found a better fix: moving the `free(reader->pattern)` from `mz_zip_reader_close` to `mz_zip_reader_delete`.
I'll update the PR soon.

Coeur:
I've moved the revert to #877.
I've added the fix for `free` here.
Now we should also look for writing a test so that this doesn't regress again.

mz_zip_rw.h
```diff
@@ -102,7 +102,7 @@ int32_t mz_zip_reader_save_all(void *handle, const char *destination_dir);
 
 /***************************************************************************/
 
-void mz_zip_reader_set_pattern(void *handle, const char *pattern, uint8_t ignore_case);
+int32_t mz_zip_reader_set_pattern(void *handle, const char *pattern, uint8_t ignore_case);
```
Coeur, line 105: If you prefer that I avoid changing the signature here, then I can add a new api named `mz_zip_reader_copy_pattern` or whatever name you want.

nmoinvaz: I don't mind it changed. Can you check to see if it needs to be updated in the `docs` folder?

Coeur:
@nmoinvaz @DreamHelium I actually have a concern with that feature request of copying the `pattern`:
- now we **must** do a copy of this `pattern`, while previously a copy wasn't necessary ; the caller of the API could have managed the lifetime of this `pattern` to be equal to the lifetime of the `reader` by themselves.
- should we do the same copy for `mz_zip_reader_set_password`? Otherwise we have inconsistent behaviours.

mz_zip_rw.h
Coeur: https://github.com/zlib-ng/minizip-ng/pull/874/commits/84a1b033d2c1bd84edaf0f68ab9f5a187aa60d87

Coeur:
In the absence of follow up from @DreamHelium, and due to inconsistency with `mz_zip_reader_set_password`, I'll close the PR.

If you have comments in favor of this feature, then let me know and I can re-open the PR.

PR #876, opened 2025-05-09
refactor: passing reader, writer, zip instead of handle

Coeur:
Coming from a different programming language, I prefer dealing with typed objects as much as possible, so I've harmonised using typed reader/writer/zip/strm where available instead of untyped handle/stream.

Also adding guards against NULL when the api is public (available in a header file).


Coeur:
@GermanAizek @nmoinvaz would you be interested in code-reviewing this refactoring? It does include a few minor fixes, like preventing a NULL dereference in public api `mz_zip_entry_seek_local_header`.

Coeur:
@nmoinvaz can we merge it? :)

Coeur:
Hi @nmoinvaz, how do we move forward here? Would you like multiple small pull requests for each guard clause for NULL pointer maybe?

nmoinvaz (approved):

PR #877, opened 2025-05-09
Revert: Make `pattern` in `mz_zip_reader` dynamically assigned (#856)

### LINKED ISSUE #873: Bug?: Extraction of a selected file
### Discussed in https://github.com/zlib-ng/minizip-ng/discussions/872

<div type='discussions-op-text'>

<sup>Originally posted by **@kabanovsan** May  9, 2025</sup>
I am using minizip (4.0.10) for the first time. It works well, but now I am facing a problem where I cannot extract a selected file. I tried in my code using the library, and the same happens if I use minizip.exe like:

cmd: `minizip.exe -x -o "C:\Users\...\PCFC.7z" "temp.prf"`
All files will be extracted.

the same happens in c++ in "minizip.c" with filename in *pattern:
https://github.com/zlib-ng/minizip-ng/blob/d265da043de2279dbe994442679e517bc69ab350/minizip.c#L55-L56

I checked in 3.0.10. In this version, it's working. So looks like it's a bug in the last version (4.0.10)
</div>

Coeur:
Alternative for: #874
Fix: #873

Undo: #856 / #855

cc @DreamHelium

This is a feature revert because:
- it's not optimal to force a copy of the `pattern` ; the caller of the API could have managed the lifetime of this `pattern` to be equal to the lifetime of the `reader` by themselves.

### DIFF
--- a/mz_zip_rw.c
+++ b/mz_zip_rw.c
@@ -39,7 +39,7 @@ typedef struct mz_zip_reader_s {
     uint16_t hash_algorithm;
     uint16_t hash_digest_size;
     mz_zip_file *file_info;
-    char *pattern;
+    const char *pattern;
     uint8_t pattern_ignore_case;
     const char *password;
     void *overwrite_userdata;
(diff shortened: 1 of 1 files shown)


Coeur:
@nmoinvaz can you prioritise this bug fix?

nmoinvaz (approved):

PR #889, opened 2025-06-16
harden mz_zip_writer_add_file against link following

peterh:
There is a small race where a file can be replaced with a link between the call to mz_os_is_symlink and the call to mz_stream_os_open. Add MZ_OPEN_MODE_NOFOLLOW to make sure that (on POSIX) a link cannot be followed.

### DIFF
--- a/mz_zip_rw.c
+++ b/mz_zip_rw.c
@@ -1629,10 +1629,13 @@ int32_t mz_zip_writer_add_file(void *handle, const char *path, const char *filen
         if (err == MZ_OK)
             file_info.linkname = link_path;
     } else if (mz_os_is_dir(path) != MZ_OK) {
+        uint32_t mode = MZ_OPEN_MODE_READ;
         stream = mz_stream_os_create();
         if (!stream)
             return MZ_STREAM_ERROR;
-        err = mz_stream_os_open(stream, path, MZ_OPEN_MODE_READ);
+        if (!writer->follow_links)
+            mode |= MZ_OPEN_MODE_NOFOLLOW;
+        err = mz_stream_os_open(stream, path, mode);
     }
 
     if (err == MZ_OK)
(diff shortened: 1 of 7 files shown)


mz_zip_rw.h
```diff
@@ -236,6 +236,8 @@ void mz_zip_writer_set_compress_level(void *handle, int16_t compress_level);
 void mz_zip_writer_set_follow_links(void *handle, uint8_t follow_links);
 /* Follow symbolic links when traversing directories and files to add */
 
+uint8_t mz_zip_writer_get_follow_links(void *handle);
```
nmoinvaz: This function needs a comment.

nmoinvaz: ~~Also lets return it in as a param as with other functions instead of in return value.~~ Nvm its fine.

peterh: > This function needs a comment.

Done. Also added it to mz_zip_rw.md, along with set_follow_links and set_store_links. Please let me know if you'd like the unrelated doc additions split out into a separate commit.

> ~Also lets return it in as a param as with other functions instead of in return value.~ Nvm its fine.

Too late, updated it before I saw your edit. Let me know if you want me to undo this change.

nmoinvaz (approved):

mz_zip_rw.c
```diff
@@ -1834,6 +1837,14 @@ void mz_zip_writer_set_follow_links(void *handle, uint8_t follow_links) {
     writer->follow_links = follow_links;
 }
 
+int32_t mz_zip_writer_get_follow_links(void *handle, uint8_t *follow_links) {
+    mz_zip_writer *writer = (mz_zip_writer *)handle;
+    if (!follow_links)
```
Coeur, line 1842:
```suggestion
    if (!writer || !follow_links)
```

peterh: This was already merged. I opened a new MR for this suggestion: #891

PR #868, opened 2025-05-01
Support any character encoding

Coeur:
The existing list of supported encodings was added by ea5e45dfe2c42ade58e260cb14b910f75b385525.
That list was quite limited.
In a more practical way, this PR supports any encoding that iconv supports:
```
➜  iconv --list | grep " [0-9]\+ "
CP037 037 EBCDIC-CP-CA EBCDIC-CP-NL EBCDIC-CP-US EBCDIC-CP-WT IBM037
CP038 038 EBCDIC-INT IBM038
CP10000 10000 CP10000_MACROMAN
CP10006 10006 CP10006_MACGREEK
CP10007 10007 CP10007_MACCYRILLIC MS-MAC-CYRILLIC
CP10029 10029 CP10029_MACLATIN2
CP1006 1006 MSCP1006
CP10079 10079 CP10079_MACICELANDIC
CP10081 10081 CP10081_MACTURKISH
CP1026 1026 IBM1026
CP1046 1046 IBM1046
CP1124 1124 IBM1124
CP1125 1125 IBM1125
CP1129 1129 IBM1129
CP1131 1131 IBM1131
CP1133 1133 IBM-CP1133 IBM1133
CP1161 1161 CSIBM1161 IBM-1161 IBM1161
CP1162 1162 CSIBM1162 IBM-1162 IBM1162 MSCP874 WINDOWS-874
CP1163 1163 CSIBM1163 IBM-1163 IBM1163
CP1250 1250 MS-EE MSCP1250 WINDOWS-1250
CP1251 1251 MS-CYRL MSCP1251 WINDOWS-1251
CP1252 1252 MS-ANSI MSCP1252 WINDOWS-1252
CP1253 1253 MS-GREEK MSCP1253 WINDOWS-1253
CP1254 1254 MS-TURK MSCP1254 WINDOWS-1254
CP1255 1255 MS-HEBR MSCP1255 WINDOWS-1255
CP1256 1256 MS-ARAB MSCP1256 WINDOWS-1256
CP1257 1257 MSCP1257 WINBALTRIM WINDOWS-1257
CP1258 1258 MSCP1258 WINDOWS-1258
CP273 273 IBM273
CP274 274 EBCDIC-BE IBM274
CP275 275 EBCDIC-BR IBM275
CP277 277 EBCDIC-CP-DK EBCDIC-CP-NO IBM277
CP278 278 EBCDIC-CP-FI EBCDIC-CP-SE IBM278
CP280 280 EBCDIC-CP-IT IBM280
[...]

PR #890, opened 2025-06-16
fix stack overrun in mz_os_utf8_string_create

peterh:
The string_encoding variable goes out of scope right after assignment to from_encoding. Referencing from_encoding after this point is technically a use-after-free error, which leads to arbitrary (undefined) behaviour.

Move string_encoding up to function scope so it survives for as long as it could be referenced.


nmoinvaz (approved):

Coeur:
Ah ah!
Thank you peterh. My bad for not seeing this! ;)
https://github.com/zlib-ng/minizip-ng/pull/868#discussion_r2070497854
Capture d’écran 2025-06-17 à 06 55 30

PR #891, opened 2025-06-17
zip_rw: validate the handle in each getter/setter

peterh:
Some of the getters and setters were making sure that the handle was non-null, and some were not. Add handle checking to every function for consistency.

---

Suggested by @Coeur in https://github.com/zlib-ng/minizip-ng/pull/889#discussion_r2151325333 . Most of the getters/setters did not check the handle previously, so maybe it's better to go the other way, remove the few checks that exist, and let invalid users crash?


nmoinvaz:
I think this needs a rebase

peterh:
> I think this needs a rebase

So it does. Looks like a few of the changes were already present in #876 (merged just now). The only conflict was on white-space.

nmoinvaz (approved):

PR #896, opened 2025-07-22
add a new example to README.md

oajlouni:
I've created a basic example to showcase minizip-ng's capabilities. This is an up-to-date example, which should hopefully be helpful to newcomers.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -265,3 +265,4 @@ Some of these may be out of date, but they can also be helpful.
 * [Compressed stream tests](https://github.com/zlib-ng/minizip-ng/blob/master/test/test_stream_compress.cc)
 * [Code to copy raw entries from one zip file to another](https://gist.github.com/chenxiaolong/bcbb0835182ef16a25f09db8d99e0619) by chenxiaolong
 * [Buffered streaming](https://gist.github.com/chenxiaolong/dbab3fbef51b9d0fa969e220dbb85967) by chenxiaolong
+* [A basic minizip-ng example](https://github.com/oajlouni/minizip-ng_example.git) by oajlouni


doc/README.md
```diff
@@ -264,4 +264,5 @@ Some of these may be out of date, but they can also be helpful.
 
 * [Compressed stream tests](https://github.com/zlib-ng/minizip-ng/blob/master/test/test_stream_compress.cc)
 * [Code to copy raw entries from one zip file to another](https://gist.github.com/chenxiaolong/bcbb0835182ef16a25f09db8d99e0619) by chenxiaolong
-* [Buffered streaming](https://gist.github.com/chenxiaolong/dbab3fbef51b9d0fa969e220dbb85967) by chenxiaolong
\ No newline at end of file
+* [Buffered streaming](https://gist.github.com/chenxiaolong/dbab3fbef51b9d0fa969e220dbb85967) by chenxiaolong
+* [A basic minizip-ng example](https://github.com/oajlouni/minizip-ng_example.git) by oajlouni
```
Coeur, line 268: "A basic example" doesn't tell _what_ is demonstrated.
Looking at your code, I see `"Usage: %s <source file> <target file> [--normal|--memory|--bufstream]\n"`, so you might want to leverage that usage description here.

In itself, the existing minizip command line (our `minizip.c`) is also an example for the api adoption. Maybe it's nice to list an alternative simpler implementation, or maybe it's redundant.

oajlouni: I'm really embarrassed to say, but I wasn't aware of minizip.c until I saw your comment 😅😅

For some reason, I was too focused on the other source files and docs that I couldn't even notice it. It could've helped me a great deal!

If my code could still be useful as an example, I can change the title to something like "Using different stream types for unzipping and copying". So, please let me know what's more suitable.

-Osama

PR #911, opened 2025-10-05
Removing superfluous const

Coeur:
Reported by @sergio-nsk in https://github.com/zlib-ng/minizip-ng/pull/834#discussion_r2302358666


nmoinvaz (approved):

tbeu:
Accoding to https://stackoverflow.com/a/38660590/8520615 the const in the declaration is superfluous while in the definition it is reasonable to avoid accidental overwriting.

Coeur:
Less than 1 upvote per year. ;)

I tend to prefer identical declarations and definitions, so I'll stick to current solution.

tbeu:
Sure, matter of taste, style and const-correctness.

PR #912, opened 2025-10-29
Rename cmake target files to avoid overwrite of PACKAGE_VERSION

bebuch:
Same patch as https://github.com/zlib-ng/zlib-ng/pull/1988 for zlib-ng.


Coeur:
One minute before the zlib-ng PR you link to was merged, another zlib-ng PR was merged with renamings and concerns by @nmoinvaz : https://github.com/zlib-ng/zlib-ng/pull/1970

So we might need to look at both zlib-ng changes to evaluate what's the next move for minizip-ng.

bebuch:
This looks like a completely distinct thing. My PRs only affect the target file names, whereas this other PR only affects the target names. Note that target file names are an implementation detail. The CMake config file API doesn't depend on them, so my change won't break anything.

nmoinvaz (approved):

nmoinvaz:
This will be commit as part of #982.

PR #913, opened 2025-10-29
fix missing includes in header files

bebuch:
Most header files are missing `stdint.h`. I copied the include mechanism from `mz.h`.

`mz_zip.h` is using `time_t` and doesn't include `time.h`.

With this changes you can just include `mz_zip.h` (and the other headers) in your own project without caring that you previously included the missing standard header files.

### DIFF
--- a/mz_crypt.h
+++ b/mz_crypt.h
@@ -11,6 +11,14 @@
 #ifndef MZ_CRYPT_H
 #define MZ_CRYPT_H
 
+#if defined(HAVE_STDINT_H)
+#  include <stdint.h>
+#elif defined(__has_include)
+#  if __has_include(<stdint.h>)
+#    include <stdint.h>
+#  endif
+#endif
+
 #ifdef __cplusplus
 extern "C" {
 #endif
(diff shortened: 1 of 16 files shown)


nmoinvaz:
You always have to include "mz.h" first.

bebuch:
Is there a reason why the other headers don't always include mz.h?

An API header that requires another API header to be manually included beforehand sounds quite error-prone. That's not what you would normally expect.

Coeur (changes requested):
Too much duplication. This is like a partial revert of fb8a1cc220008546f9e2baad48bfb71095118dc6 from November 20, 2018.

Maybe replace your snippets with `#include "mz.h"`, but even then I'm not sure if the change would be approved: the `mz_*.h` weren't made to be used self-sufficient, and `mz.h` has been a requirement since its introduction on October 16, 2017 in 00cca9f5114d29f5a9f2fed8e70b1400ed4679a4 (aka minizip 2.0 - nowadays we're at minizip 4.0). And even before it, there was its ancestor `mz_error.h` which might have been already required.

nmoinvaz:
Not all those headers are intended to be fool-proof when including. Sorry, I don't want to increase the complexity at this time. I do have example code such as minigzip that should show the proper way to use the files.

PR #917, opened 2025-10-30
`mz_stream_zstd_set_prop_int64`: Implicit conversion loses integer precision: 'int64_t' (aka 'long long') to 'int'

Coeur:
Warning since #888 by @peterh 

Image

Solution: same as line 306, we make the conversion explicit with a cast.


nmoinvaz (approved):

peterh:
Good catch, thanks. LGTM

PR #925, opened 2025-11-08
Remove `build-src-dir` in Actions

Coeur:
Issue: when a fork is named differently than strictly "minizip-ng", then `Ubuntu GCC OSB` is failing as "../minizip-ng" doesn't exist. This causes systematic build failures on my fork:
https://github.com/Coeur/minizip/actions/runs/19191333603/job/54866172824

Solution: no need to hardcode `build-src-dir` at all.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -54,12 +54,12 @@ jobs:
             cmake-args: -D MZ_CODE_COVERAGE=ON
             codecov: ubuntu_gcc
 
+          # out-of-source build
           - name: Ubuntu GCC OSB
             os: ubuntu-latest
             compiler: gcc
             cxx-compiler: g++
             build-dir: ../build
-            build-src-dir: ../minizip-ng
 
           - name: Ubuntu Clang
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -276,7 +276,7 @@ jobs:
         token: ${{ secrets.CODECOV_TOKEN }}
         flags: ${{ matrix.codecov }}
         name: ${{ matrix.name }}
-        directory: ${{ matrix.build-src-dir || '.' }}
+        directory: .
```
nmoinvaz, line 279: I wonder if this needs to be a list of `${{ matrix.build-src-dir || '.' }}` and `.`.

Coeur: I do not understand your question. I'm removing `matrix.build-src-dir` in this PR.

nmoinvaz: Sorry, I think I was confused.

nmoinvaz (approved):

PR #929, opened 2025-11-21
fix(cmake): Correct Zlib-ng static link by separating link and install targets

Chirimen-Jako:
When building minizip-ng as a static library (undefined BUILD_SHARED_LIBS) with Zlib-ng, the .vcxproj generated by CMake incorrectly referred to the dynamic library target (`zlib`) instead of the static library target (`zlibstatic`) in its linker settings.

This fix separates the targets in CMake:
- Use `zlibstatic` for target_link_libraries() to ensure the generated .vcxproj correctly references the static library.
- Use `zlib` for install(TARGETS...) to prevent CMake alias errors during installation (CI).


CMakeLists.txt
```diff
@@ -721,7 +723,17 @@ if(MZ_LZMA)
     set_target_properties(${MINIZIP_TARGET} PROPERTIES C_STANDARD 99)
 endif()
 
-target_link_libraries(${MINIZIP_TARGET} PUBLIC ${MINIZIP_LIB} ${MINIZIP_DEP})
+set(MINIZIP_LINK_DEP ${MINIZIP_DEP})
+
+if(MINIZIP_ZLIB_DEP)
+    list(INSERT MINIZIP_DEP 0 ${MINIZIP_ZLIB_DEP})
```
nmoinvaz, line 729: Why get the first item from MINIZIP_DEP? I wonder if there a better way to do this without doing this.

nmoinvaz:
AFAIK, this should no longer be needed. Let me know if that is not the case. Thanks!

PR #939, opened 2026-01-19
Add some files to .gitignore

pmqs:
Update .gitignore to include these license files
```
LICENSE.dec.aes
LICENSE.dec.pkcrypt
LICENSE.enc.aes
LICENSE.enc.pkcrypt
```

`test_compat.cc` doesn't delete `compat.zip` on completion - change gets the test harness to delete it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,9 @@ third_party/
 
 # IDE
 .vscode
+
+# License files
+LICENSE.dec.aes
+LICENSE.dec.pkcrypt
+LICENSE.enc.aes
+LICENSE.enc.pkcrypt
--- a/.gitignore
+++ b/.gitignore
@@ -15,5 +15,5 @@ LICENSE.dec.pkcrypt
 LICENSE.enc.aes
 LICENSE.enc.pkcrypt
 
-# test/test_compat.cc can leave compat.zip handing around
+# test/test_compat.cc can leave compat.zip hanging around
 compat.zip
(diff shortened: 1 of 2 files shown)


test/test_compat.cc
```diff
@@ -289,4 +289,13 @@ TEST(compat, unzip64) {
     test_unzip_compat(unzip);
     unzClose(unzip);
 }
+
+/*  Keep this dummy test last in this file.
+    It's purpose is to delete "compat.zip" after all other tests have run.
+*/
+
+TEST(compat, final) {
+    if (mz_os_file_exists("compat.zip") != MZ_OK)
```
nmoinvaz, line 298: What about just removing this file on the next time the tests are run, like where it is created, delete before checking?

pmqs: > What about just removing this file on the next time the tests are run, like where it is created, delete before checking?

The point of deleting the file at the end is so that git doesn't see it as an untracked file, like the LICENSE.* files.

Other option is to just include it in `.gitignore`

pmqs:
Added `compat.zip` to `.gitignore` in addition to getting `test/test_compat.cc` to delete it. Covers the use case where `test/test_compat.cc`  fails mid-stream and doesn't get the chance to delete it.

Guards against people (like me) accidentally checked in if it ends up in `git status` land.

.gitignore
```diff
@@ -14,3 +14,6 @@ LICENSE.dec.aes
 LICENSE.dec.pkcrypt
 LICENSE.enc.aes
 LICENSE.enc.pkcrypt
+
+# test/test_compat.cc can leave compat.zip handing around
```
Coeur: I'm not a native speaker, but shouldn't it be "hanging"?

pmqs: > I'm not a native speaker, but shouldn't it be "hanging"?

It should indeed!

Unfortunately I don't have any excuse :-)

PR #949, opened 2026-02-02
build: avoid MIN macro for Windows

### LINKED ISSUE #948: minizip-ng build fails on Windows with MinGW64 due to missing MIN macro
minizip-ng fails to build with MinGW64 starting with commit 5880418, complaining of implicit declaration of MIN.

This commit added mz_strm_ppmd.c which contains

`#include <sys/param.h>  // for MIN()`

However, MinGW64 headers don't contain macros for MIN/MAX. An inline definition may be needed.

Coeur:
Fix #948.

While we could maybe add some `#import <windows.h>` or create our own macros, I feel the overhead is not needed for a single usage of MIN.

I use the ternary operator because it's already in use the line just below (`restor =`).

I also fix the documentation which had some outdated values. We know we're between PPMD8_MIN_ORDER (2) and PPMD8_MAX_ORDER (16), so the rest of the documentation should be a mem_size between `1 << 2-1` (2) and `1 << 8-1` (128).

@pmqs

### DIFF
--- a/mz_strm_ppmd.c
+++ b/mz_strm_ppmd.c
@@ -8,7 +8,6 @@
    See the accompanying LICENSE file for the full text of the license.
 */
 
-#include <sys/param.h>  // for MIN()
 #include <assert.h>
 
 #include "mz.h"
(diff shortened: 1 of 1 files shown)


Coeur:
@nmoinvaz this PR should be easy, replacing a `MIN()` with `< ? :`.

nmoinvaz (approved):

PR #930, opened 2025-11-28
Fix modified_date when zipping

Coeur:
This is an experimental attempt to fix #921 by reverting one of the lines modified by https://github.com/zlib-ng/minizip-ng/commit/7df56f7cc8438b1b67c881cad049b29bf075bccd.

This fix is untested on my end, and purely based on code review.

PR #951, opened 2026-02-02
Document that mz_zip_tm_to_time_t requires a year-1900 indexed time struct.

Coeur:
This is a documentation followup of #930 (and #820).

I wanted to better understand why my experimental fix was working and so I explored the possible values for a dos date, a tm struct and a time_t.

For edge cases:
- a `dos date` is **year 1980 indexed**. A dos date of 0 is around December 1979 due to months being -1 indexed. It doesn't take negative values. Maximum value is dos year 127, so that's roughly year 2108.
- a `tm struct` can have multiple meanings depending on the interpretation of the year ; but as an argument of `mktime`, it's considered to be **year 1900 indexed**. It accepts negative values and as such may go 2 billion years in the past and the future.
- a `time_t` is **year 1970 indexed**. It accepts negative values and can go pretty far both directions.

Now, as I wrote here, a `tm struct` can have multiple meanings depending on the interpretation of the year, and that's why in _some contexts_, it's considered to be **year 0 indexed** (example, the value was 2025). And if you're in such context and want to pass it to `mktime`, then yes, you must first subtract 1900 to the year (here it becomes 125) to convert the struct to be **year 1900 indexed**. But if you were already in the context of being year 1900 indexed (example, the value was 125), then subtracting 1900 (here it becomes -1775) will get you into... **year 3800 indexed**. Which will produce a very negative value of `time_t` (it will produce a time_t equivalent to year 1900-1775 = 125). And since those dates are way too old for the operating system, it will replace them with whatever is the minimal supported date, which was 1980 for @claraSophos.

So what did my fix do?
Well, it passes the `tm struct` to `mz_zip_tm_to_dosdate` where the magic operates: by looking at the range of tm_year and normalising it into `fixed_tm`, it will always produce a valid year-1980 indexed dos date, no matter what was the year origin of the given parameter. Subsequently, it converts it back to a year-1900 indexed time struct, which can finally be passed safely to `mktime`.

So what does the present PR do? I'm simply adding comments saying we require a "year-1900 indexed" parameter for mktime, and that `mz_zip_dosdate_to_raw_tm` just so happen return a "year-1900 indexed" time.

I also made the compat tests compatible with HAVE_LIBCOMP (which is "lib compatibility") where they make the most sense to be tested, and which helped me in this situation.

### DIFF
--- a/mz_zip.c
+++ b/mz_zip.c
@@ -2669,24 +2669,31 @@ int32_t mz_zip_extrafield_write(void *stream, uint16_t type, uint16_t length) {
 
 /***************************************************************************/
 
+/// This is mostly validating a given time struct is convertible to and from a valid dos date.
+/// This does not however perform a calendar validation (like February 31).
 static int32_t mz_zip_invalid_date(const struct tm *ptm) {
 #define datevalue_in_range(min, max, value) ((min) <= (value) && (value) <= (max))
     return (!datevalue_in_range(0, 127 + 80, ptm->tm_year) || /* 1980-based year, allow 80 extra */
-            !datevalue_in_range(0, 11, ptm->tm_mon) || !datevalue_in_range(1, 31, ptm->tm_mday) ||
-            !datevalue_in_range(0, 23, ptm->tm_hour) || !datevalue_in_range(0, 59, ptm->tm_min) ||
-            !datevalue_in_range(0, 59, ptm->tm_sec));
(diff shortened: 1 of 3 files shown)


nmoinvaz:
Can you switch to use multi-line comments?

nmoinvaz:
Other than that, it looks good, thanks!

nmoinvaz (approved):

nmoinvaz:
Thanks @Coeur!

PR #957, opened 2026-02-04
Compiler warning: use of `mktemp` is dangerous. Fixes #956

### LINKED ISSUE #956: Compiler warning: use of `mktemp` is dangerous
Just looking at the GitHub workflow output for gcc & clang on Ubuntu, I see

```
[ 57%] Linking C executable minigzip
/usr/bin/ld: libminizip.a(mz_os_posix.c.o): in function `mz_os_get_temp_path':
/home/runner/work/minizip-ng/minizip-ng/mz_os_posix.c:374:(.text+0x1023): warning: the use of `mktemp' is dangerous, better use `mkstemp' or `mkdtemp'
[ 57%] Built target minigzip_cli
[ 60%] Building C object CMakeFiles/minizip_cli.dir/minizip.c.o
[ 62%] Linking C executable minizip
/usr/bin/ld: libminizip.a(mz_os_posix.c.o): in function `mz_os_get_temp_path':
/home/runner/work/minizip-ng/minizip-ng/mz_os_posix.c:374:(.text+0x1023): warning: the use of `mktemp' is dangerous, better use `mkstemp' or `mkdtemp'
```

Adding `-Werror` to the build would catch this type of thing automatically. Not sure if that is what you want here.

pmqs:
Silence the compiler warning below by using `mkdtemp` instead of `mktemp`
```
57%] Linking C executable minigzip
/usr/bin/ld: libminizip.a(mz_os_posix.c.o): in function `mz_os_get_temp_path':
/home/runner/work/minizip-ng/minizip-ng/mz_os_posix.c:374:(.text+0x1023): warning: the use of `mktemp' is dangerous, better use `mkstemp' or `mkdtemp'
[ 57%] Built target minigzip_cli
[ 60%] Building C object CMakeFiles/minizip_cli.dir/minizip.c.o
[ 62%] Linking C executable minizip
/usr/bin/ld: libminizip.a(mz_os_posix.c.o): in function `mz_os_get_temp_path':
/home/runner/work/minizip-ng/minizip-ng/mz_os_posix.c:374:(.text+0x1023): warning: the use of `mktemp' is dangerous, better use `mkstemp' or `mkdtemp'
```

### DIFF
--- a/mz_os_posix.c
+++ b/mz_os_posix.c
@@ -383,7 +383,7 @@ int32_t mz_os_get_temp_path(char *path, int32_t max_path, const char *prefix) {
     if (result < 0 || result >= max_path)
         return MZ_BUF_ERROR;
 
-    free(temp_path) ;
+    free(temp_path);
     return MZ_OK;
 }
 
--- a/mz_os_posix.c
+++ b/mz_os_posix.c
@@ -389,7 +389,7 @@ int32_t mz_os_get_temp_path(char *path, int32_t max_path, const char *prefix) {
     /* Use current time for the filename                */
     result = snprintf(path, max_path, "%s/%lux", temp_path, time(NULL));
     if (result < 0 || result >= max_path) {
-        rmdir(temp_path);
+        rmdir(temp_path);  // clean up the created directory
         free(temp_path);
         return MZ_BUF_ERROR;
     }
--- a/mz_os_posix.c
+++ b/mz_os_posix.c
@@ -352,8 +352,8 @@ int32_t mz_os_read_symlink(const char *path, char *target_path, int32_t max_targ
 
 int32_t mz_os_get_temp_path(char *path, int32_t max_path, const char *prefix) {
     const char *tmp_dir = NULL;
+    char *temp_path = (char *)calloc(max_path, sizeof(char));
     int32_t result = 0;
-    char temp_path[max_path];
 
     if (!path || max_path <= 0)
         return MZ_PARAM_ERROR;
--- a/mz_os_posix.c
+++ b/mz_os_posix.c
@@ -383,6 +383,7 @@ int32_t mz_os_get_temp_path(char *path, int32_t max_path, const char *prefix) {
     if (result < 0 || result >= max_path)
         return MZ_BUF_ERROR;
 
+    free(temp_path) ;
     return MZ_OK;
 }
 
--- a/mz_os_posix.c
+++ b/mz_os_posix.c
@@ -378,7 +378,8 @@ int32_t mz_os_get_temp_path(char *path, int32_t max_path, const char *prefix) {
         return MZ_INTERNAL_ERROR;
 
     /* Create a filename inside the temporary directory */
-    result = snprintf(path, max_path, "%s/f", temp_path);
+    /* Use current time for the filename                */
+    result = snprintf(path, max_path, "%s/%lux", temp_path, time(NULL));
     if (result < 0 || result >= max_path)
         return MZ_BUF_ERROR;
 
--- a/mz_os_posix.c
+++ b/mz_os_posix.c
@@ -366,10 +366,6 @@ int32_t mz_os_get_temp_path(char *path, int32_t max_path, const char *prefix) {
     if (!tmp_dir)
         tmp_dir = "/tmp";
 
-    /* Check for no environment variable set at all */
-    if (!tmp_dir)
-        return MZ_INTERNAL_ERROR;
-
     /* Build template path for mkdtemp: <tmp_dir>/<prefix>XXXXXX */
     result = snprintf(temp_path, max_path, "%s/%sXXXXXX", tmp_dir, prefix ? prefix : "");
     if (result < 0 || result >= max_path)
--- a/mz_os_posix.c
+++ b/mz_os_posix.c
@@ -352,7 +352,7 @@ int32_t mz_os_read_symlink(const char *path, char *target_path, int32_t max_targ
 
 int32_t mz_os_get_temp_path(char *path, int32_t max_path, const char *prefix) {
     const char *tmp_dir = NULL;
-    char *temp_path = (char *)calloc(max_path, sizeof(char));
+    char *temp_path;
     int32_t result = 0;
 
     if (!path || max_path <= 0)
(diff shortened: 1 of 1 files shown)


mz_os_posix.c
```diff
@@ -365,15 +366,26 @@ int32_t mz_os_get_temp_path(char *path, int32_t max_path, const char *prefix) {
...
-    /* mktemp replaces XXXXXX with unique characters */
-    if (!mktemp(path))
+    /* Create a temporary directory.
+       mkdtemp replaces XXXXXX with unique characters
+    */
+    if (!mkdtemp(temp_path))
         return MZ_INTERNAL_ERROR;
 
+    /* Create a filename inside the temporary directory */
+    result = snprintf(path, max_path, "%s/f", temp_path);
```
nmoinvaz: The filename is f?

pmqs: > The filename is f?

Yes.

mz_os_posix.c
```diff
@@ -365,15 +366,26 @@ int32_t mz_os_get_temp_path(char *path, int32_t max_path, const char *prefix) {
     if (!tmp_dir)
         tmp_dir = "/tmp";
 
-    /* Build template path for mktemp: <tmp_dir>/<prefix>XXXXXX */
-    result = snprintf(path, max_path, "%s/%sXXXXXX", tmp_dir, prefix ? prefix : "");
+    /* Check for no environment variable set at all */
+    if (!tmp_dir)
+        return MZ_INTERNAL_ERROR;
+
```
tbeu:
```suggestion
```

Is this reachable actually?

pmqs: Good catch. I missed the last if statement in  the  block of code before this.  I'll update it


```C
    if (!tmp_dir)
        tmp_dir = "/tmp";
```

nmoinvaz:
Maybe I am a dumb-dumb, but why is random dir with a fixed name more secure than a random filename with a fixed dir?

pmqs:
> Maybe I am a dumb-dumb, but why is random dir with a fixed name more secure than a random filename with a fixed dir?

I believe that `mktemp` is deprecated because it doesn't create a the temporary file in the filesystem, so there is a potential race condition where a bad actor guesses the filename before you get a change to use it.

`mkdtemp` atomically creates the directory, so no race condition.

Coeur:
>BUGS
>
>       Never use mktemp().  Some implementations follow 4.3BSD and
>       replace XXXXXX by the current process ID and a single letter, so
>       that at most 26 different names can be returned.  Since on the one
>       hand the names are easy to guess, and on the other hand there is a
>       race between testing whether the name exists and opening the file,
>       every use of mktemp() is a security risk.  The race is avoided by
>       [mkstemp(3)](https://man7.org/linux/man-pages/man3/mkstemp.3.html) and [mkdtemp(3)](https://man7.org/linux/man-pages/man3/mkdtemp.3.html).

Source: https://man7.org/linux/man-pages/man3/mktemp.3.html

nmoinvaz:
@Coeur thanks that makes sense.

I suggest instead of `f` for filename we use `snprintf` with just the unix timestamp.

pmqs:
> @Coeur thanks that makes sense.
> 
> I suggest instead of `f` for filename we use `snprintf` with just the unix timestamp.

Done

mz_os_posix.c
```diff
@@ -353,6 +353,7 @@ int32_t mz_os_read_symlink(const char *path, char *target_path, int32_t max_targ
 int32_t mz_os_get_temp_path(char *path, int32_t max_path, const char *prefix) {
     const char *tmp_dir = NULL;
     int32_t result = 0;
+    char temp_path[max_path];
```
nmoinvaz: Sorry for this, but I think this is a newer C standard thing right? Maybe just best to do calloc().

tbeu: C99 brought stack-allocated variable-length arrays, see https://en.wikipedia.org/wiki/Variable-length_array#C.

Yes, better stick with heap-allocated array here.

pmqs: Ha!  I knew that zlib-ng policed this one in cmake and assumed minizip-ng did as well.

I've made the change, but can we get cmake to police a minimum version?

mz_os_posix.c
```diff
@@ -352,6 +352,7 @@ int32_t mz_os_read_symlink(const char *path, char *target_path, int32_t max_targ
 
 int32_t mz_os_get_temp_path(char *path, int32_t max_path, const char *prefix) {
     const char *tmp_dir = NULL;
+    char *temp_path = (char *)calloc(max_path, sizeof(char));
```
nmoinvaz:
```
char *temp_path = NULL;
```

mz_os_posix.c
```diff
@@ -365,15 +366,23 @@ int32_t mz_os_get_temp_path(char *path, int32_t max_path, const char *prefix) {
     if (!tmp_dir)
         tmp_dir = "/tmp";
 
-    /* Build template path for mktemp: <tmp_dir>/<prefix>XXXXXX */
-    result = snprintf(path, max_path, "%s/%sXXXXXX", tmp_dir, prefix ? prefix : "");
+    /* Build template path for mkdtemp: <tmp_dir>/<prefix>XXXXXX */
```
nmoinvaz, line 369:
```
temp_path = (char *)calloc(max_path, sizeof(char));
```
Don't forget to do `free(temp_path)` also for every failure path after.

nmoinvaz (changes requested):

mz_os_posix.c
pmqs: > ```
> temp_path = (char *)calloc(max_path, sizeof(char));
> ```
> 
> Don't forget to do `free(temp_path)` also for every failure path after.

Wow! I run with address sanitizer enabled all the time to catch things like this.

Hmm, just looking at the compiler commandline I'm not seeing the expected options.

Anyway, I've added the `free`

nmoinvaz: This is still not fixed, there are return paths where it would leak memory.

pmqs: I'm on it.

pmqs:
@nmoinvaz think I've covered all the review comments. Ping if I've missed anything

nmoinvaz:
Rebased.

pmqs:
> Rebased.

Thanks @nmoinvaz -- you should have got me to do a force push in the PR.

mz_os_posix.c
nmoinvaz: `temp_path = (char *)calloc(max_path, sizeof(char));` needs to be set close to be using, otherwise it will leak on early return.

pmqs:
Hmm, git/github don't seem to want to let me do a force-push on the latest updates.

PR #960, opened 2026-02-10
Harden the cmake Address Sanitizer detection code.

pmqs:
Currently the sanitizer detection code will silently ignore error cases -- like when given an unknown option or when it is asked for a valid sanitizer that isn't available for the platform. This one has stung me a few times.

This change makes it a fatal error to
1. ask for an address sanitizer option that doesn't exist
2. asks for a valid option that the current platform doesn't support.

One thng that dropped out with this change is the use of the Memory Sanitizer in `build.yml`, Turns out that it is not available on Ubuntu, so has been adding nothing to our test coverage. I think that may be a limitation of virtualisation. I've just disabled it for now.

This is what `build.yml` had when (not) detecting the memory Sanitizer:
```
-- Checking for Address Sanitizer Option
-- Performing Test HAVE_SANITIZER_memory
-- Performing Test HAVE_SANITIZER_memory - Failed
-- Memory sanitizer is not supported
```


nmoinvaz:
The sanitizer CMake is too borrowed from zlib-ng, so I'm uncomfortable changing it.

CMakeLists.txt
```diff
@@ -691,14 +692,31 @@ if(MZ_COMPAT)
...
+    elseif(MZ_SANITIZER_LOWER STREQUAL "memory")
+        add_memory_sanitizer("MZ_STATUS")
+    elseif(MZ_SANITIZER_LOWER STREQUAL "thread")
+        add_thread_sanitizer("MZ_STATUS")
+    elseif(MZ_SANITIZER_LOWER STREQUAL "undefined")
+        add_undefined_sanitizer("MZ_STATUS")
+    endif()
+    if(NOT MZ_STATUS)
+        set(MZ_SANITIZER_ENABLED OFF)
+        message(FATAL_ERROR "Cannot enable the Address Sanitizer '${MZ_SANITIZER}' option")
```
Coeur: Let's just log "Sanitizer" without "Address", as it can be any of them.

CMakeLists.txt
```diff
@@ -691,14 +692,31 @@ if(MZ_COMPAT)
...
+    endif()
+elseif(MZ_SANITIZER_LOWER STREQUAL "none")
+    set(MZ_SANITIZER_ENABLED OFF)
+    message(STATUS " Address Sanitizer not enabled")
```
Coeur: Let's just log "Sanitizer" without "Address", as it can be any of them.

CMakeLists.txt
```diff
@@ -1083,5 +1101,6 @@ add_feature_info(MZ_BUILD_TESTS MZ_BUILD_TESTS "Builds minizip test executable")
 add_feature_info(MZ_BUILD_UNIT_TESTS MZ_BUILD_UNIT_TESTS "Builds minizip unit test project")
 add_feature_info(MZ_BUILD_FUZZ_TESTS MZ_BUILD_FUZZ_TESTS "Builds minizip fuzzer executables")
 add_feature_info(MZ_CODE_COVERAGE MZ_CODE_COVERAGE "Builds with code coverage flags")
+add_feature_info(MZ_SANITIZER MZ_SANITIZER_ENABLED "Builds with Address Sanitizer")
```
Coeur: Let's just log "Sanitizer Support" instead of "Address Sanitizer", as it can be any of them.

Coeur (changes requested):

Coeur:
> The sanitizer CMake is too borrowed from zlib-ng, so I'm uncomfortable changing it.

The change looks good to me.

@pmqs can you also offer a PR for the zlib-ng repository eventually?

CMakeLists.txt
pmqs: Done

CMakeLists.txt
pmqs: Done

CMakeLists.txt
pmqs: Done

pmqs: Done

pmqs:
> > The sanitizer CMake is too borrowed from zlib-ng, so I'm uncomfortable changing it.
> 
> The change looks good to me.
> 
> @pmqs can you also offer a PR for the zlib-ng repository eventually?

@Coeur, as it happens I'm ahead of you. Already looking at the equivalent change in zlib-ng because there is a knock-on impact with it for Sanitizer support with the Microsoft C compiler. Will write up the details in the zlib-ng PR.

I'll park this change for now & get the same fix done in the much simpler zlib-ng use case sorted first.

CMakeLists.txt
```diff
@@ -691,14 +692,31 @@ if(MZ_COMPAT)
...
 # Detect available sanitizers
-if(MZ_SANITIZER STREQUAL "Address")
-    add_address_sanitizer()
-elseif(MZ_SANITIZER STREQUAL "Memory")
-    add_memory_sanitizer()
-elseif(MZ_SANITIZER STREQUAL "Thread")
-    add_thread_sanitizer()
-elseif(MZ_SANITIZER STREQUAL "Undefined")
-    add_undefined_sanitizer()
+message(STATUS "Checking for Address Sanitizer Option")
```
Coeur: maybe remove "Address" here?

Coeur (approved):

.github/workflows/build.yml
```diff
@@ -25,12 +25,13 @@ jobs:
             cmake-args: -D MZ_SANITIZER=Undefined
             codecov: ubuntu_gcc_undefined
 
-          - name: Ubuntu GCC MSAN
-            os: ubuntu-latest
-            compiler: gcc
-            cxx-compiler: g++
-            cmake-args: -D MZ_SANITIZER=Memory
-            codecov: ubuntu_gcc_msan
+          # Memory Sanitizer not currenly supported on GitHub
```
nmoinvaz: https://stackoverflow.com/questions/71675493/how-use-memory-sanitizer-with-use-gcc
From what I have seen, MSAN is not supported with GCC, so I think this can be removed.

nmoinvaz: https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html
No `--fsanitize=memory`.

pmqs: Yep, that's my understanding as well -- clang only

nmoinvaz:
Has this been updated with the latest changes from zlib-ng?

pmqs:
> Has this been updated with the latest changes from zlib-ng?

Yes, it has.

nmoinvaz:
@pmqs can you please rebase this and see if those Windows builds are fixed? I would do it myself, but last time it messed it up for you.

pmqs:
> @pmqs can you please rebase this and see if those Windows builds are fixed? I would do it myself, but last time it messed it up for you.

@nmoinvaz try that now

nmoinvaz:
Rather than deleting the MSAN job, change it to use clang, matching zlib-ng's approach:                                                                                                    
                                                                          
  - rename to "Ubuntu Clang MSAN"                                                                                                                                                            
  - set compiler: clang-20 / cxx-compiler: clang++-20                     
  - in the apt install step, add libclang-rt-20-dev (provides the MSAN runtime) — without it, link fails even though compile succeeds

cmake/detect-sanitizer.cmake
```diff
@@ -55,11 +55,16 @@ macro(add_address_sanitizer)
...
         message(STATUS "Address sanitizer is enabled: ${supported_checks}")
-        add_compile_options(-fsanitize=${supported_checks})
-        add_link_options(-fsanitize=${supported_checks})
+        if(NOT MSVC)
+            add_compile_options("-fsanitize=${supported_checks}")
+            add_link_options("-fsanitize=${supported_checks}")
+        else()
+            add_compile_options("/fsanitize=${supported_checks}" "/Zi")
+            # Note that the MSVC linker doesn't use /fsanitizer
+        endif()
```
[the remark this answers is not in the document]
nmoinvaz: `check_sanitizer_support` should probe which flags are supported and MSVC should return only `address` in `supported_checks`.

nmoinvaz (approved):

nmoinvaz:
@pmqs thanks for your help! I will fix the MSAN issues from here.

pmqs:
> @pmqs thanks for your help! I will fix the MSAN issues from here.

No worries @nmoinvaz . My fix was a work-on-progress that I didn't get a chance to look at  yesterday. Thanks for finishing it off.

PR #976, opened 2026-04-09
Fix two bugs causing premature exiting the PPMd decompression

solbjorn:
1. PPMd doesn't use a buffer to read input data, it reads data byte-by-byte, and it doesn't have a loop to process the whole requested size buffer-by-buffer, but still has a cap of maximum input bytes set to `INT16_MAX`. As a result, the decompression function never decodes more bytes than this number and exits.
2. The byte-by-byte decoding loop decrements `avail_in` together with `avail_out`, but 1 byte of compressed data doesn't mean 1 byte of compressed data. This results in that the decompression ends after writing `avail_in` of **decompressed** data, which is always less than the requested size.

Fix the first one by removing the cap of `INT16_MAX` and using the real boundaries (`max_avail_in`) capped to `INT32_MAX` just in case (the requested size is of `int32`, but `max_avail_in` is of `int64_t` and can be bigger than the former).
Fix the second one by tracking how many **compressed** bytes were actually read and exit only when the actual limit is exhausted.

This removes the need of calling `mz_zip{,_reader}_entry_read()` in a loop when the ZIP is compressed with PPMd and the caller wants to read the whole file or a chunk larger than the compressed size (or `INT16_MAX`).

+ a bonus from CodeRabbit to make sure the LZMA SDK code won't read past the limit.

Tested on Windows w/clang-cl and a PPMd-compressed ZIP containing up to 700 Kb -sized files.

### DIFF
--- a/mz_strm_ppmd.c
+++ b/mz_strm_ppmd.c
@@ -309,18 +309,16 @@ int32_t mz_stream_ppmd_read(void *stream, void *buf, int32_t size) {
     mz_stream_ppmd *ppmd = (mz_stream_ppmd *)stream;
     uint8_t *next_out = buf;
     int32_t avail_out;
-    int32_t avail_in = sizeof(ppmd->buffer);
+    int64_t start_in = ppmd->total_in;
+    int64_t avail_in = (ppmd->max_total_in > 0 ? ppmd->max_total_in : INT64_MAX) - start_in;
     int sym = 0;
     int32_t written = 0;
 
     if (ppmd->end_stream)
         return MZ_OK;
 
-    if (ppmd->max_total_in > 0 && avail_in > (ppmd->max_total_in - ppmd->total_in))
-        avail_in = (int32_t)(ppmd->max_total_in - ppmd->total_in);
-
     /* Decode input to fill the output buffer. */
-    for (avail_out = size; avail_out > 0 && avail_in > 0; avail_out--, avail_in--) {
+    for (avail_out = size; avail_out > 0 && avail_in > (ppmd->total_in - start_in); avail_out--) {
         sym = Ppmd8_DecodeSymbol(&ppmd->ppmd8);
 
         /* There are two ways to terminate the loop early:
--- a/mz_strm_ppmd.c
+++ b/mz_strm_ppmd.c
@@ -150,6 +150,11 @@ static Byte reader(void *p) {
     uint8_t b;
     int32_t status;
 
+    if (ppmd->max_total_in > 0 && ppmd->total_in >= ppmd->max_total_in) {
+        ppmd->error = MZ_STREAM_ERROR;
+        return 0;
+    }
+
     if ((status = mz_stream_read_uint8((mz_stream_ppmd *)ppmd->stream.base, &b)) != MZ_OK) {
         ppmd->error = status;
         b = 0;


mz_strm_ppmd.c
```diff
@@ -310,17 +310,18 @@ int32_t mz_stream_ppmd_read(void *stream, void *buf, int32_t size) {
     uint8_t *next_out = buf;
     int32_t avail_out;
     int32_t avail_in = sizeof(ppmd->buffer);
```
nmoinvaz:
```suggestion
int64_t limit = ppmd->max_total_in ? ppmd->max_total_in : INT64_MAX;                                                                                                                         
int64_t avail_in = cap - start_in; 
```
Then we don't need the first modified if statement.

mz_strm_ppmd.c
```diff
@@ -310,17 +310,18 @@ int32_t mz_stream_ppmd_read(void *stream, void *buf, int32_t size) {
...
+    int64_t start_in = ppmd->total_in;
     int sym = 0;
     int32_t written = 0;
 
     if (ppmd->end_stream)
         return MZ_OK;
 
-    if (ppmd->max_total_in > 0 && avail_in > (ppmd->max_total_in - ppmd->total_in))
-        avail_in = (int32_t)(ppmd->max_total_in - ppmd->total_in);
+    if (ppmd->max_total_in > start_in && (ppmd->max_total_in - start_in) <= INT32_MAX)
```
nmoinvaz: `ppmd->max_total_in > start_in`
case was not handled when `==`, it would do `INT32_MAX` when it should have been zero I think.

nmoinvaz (changes requested):

solbjorn:
@nmoinvaz, makes sense, done. The initial version was sorta half-way, dunno what I was thinking of :D

solbjorn:
Re the AI feedback: should I address it? The byte reader is not changed in this PR, so this "potential OOB" could happen before as well. Plus IIRC PPMd never reads more than 1 byte at a time and with all our checks inside the loop declaration, this just can't happen.

mz_strm_ppmd.c
```diff
@@ -309,18 +309,16 @@ int32_t mz_stream_ppmd_read(void *stream, void *buf, int32_t size) {
     mz_stream_ppmd *ppmd = (mz_stream_ppmd *)stream;
     uint8_t *next_out = buf;
     int32_t avail_out;
-    int32_t avail_in = sizeof(ppmd->buffer);
+    int64_t start_in = ppmd->total_in;
+    int64_t avail_in = (ppmd->max_total_in > 0 ? ppmd->max_total_in : INT64_MAX) - start_in;
```
nmoinvaz, line 318: This misses again when ppmd->max_total_in == 0. I recommend just using the code I suggested.

nmoinvaz (changes requested):

mz_strm_ppmd.c
```diff
@@ -309,18 +309,16 @@ int32_t mz_stream_ppmd_read(void *stream, void *buf, int32_t size) {
...
 
     if (ppmd->end_stream)
         return MZ_OK;
 
-    if (ppmd->max_total_in > 0 && avail_in > (ppmd->max_total_in - ppmd->total_in))
-        avail_in = (int32_t)(ppmd->max_total_in - ppmd->total_in);
-
     /* Decode input to fill the output buffer. */
-    for (avail_out = size; avail_out > 0 && avail_in > 0; avail_out--, avail_in--) {
+    for (avail_out = size; avail_out > 0 && avail_in > (ppmd->total_in - start_in); avail_out--) {
```
nmoinvaz, line 326: @coderabbitai, that doesn't seem to prevent the OOB read though right? What is the max number of bytes that PPmd can read in one decodeSymbol?

nmoinvaz:
> Re the AI feedback: should I address it? The byte reader is not changed in this PR, so this "potential OOB" could happen before as well. Plus IIRC PPMd never reads more than 1 byte at a time and with all our checks inside the loop declaration, this just can't happen.

Yes let's add that as a separate commit. Use the AI's suggested fix.

mz_strm_ppmd.c
solbjorn: Your suggestion:
```c
int64_t limit = ppmd->max_total_in ? ppmd->max_total_in : INT64_MAX;                                                                                                                         
int64_t avail_in = cap - start_in; 
```
How is this different?

Other decompressors have the same logic, except that they take `sizeof(buffer)` instead of `INT64_MAX` when `max_total_in == 0`.

The bot's suggestion also does bounds checking only if `max_total_in > 0`.

solbjorn:
> Yes let's add that as a separate commit. Use the AI's suggested fix.

No problem. Just pls tell me how you'd prefer me to commit it: I guess not with `--author="CodeRabbit"`, rather mentioning it in a `Suggested-by:` (`Assisted-by:`?)?

mz_strm_ppmd.c
nmoinvaz: `ppmd->max_total_in ? ` is `>= 0`.

I checked the PR originally with Claude, and that is when it mentioned if it was `== 0` then `avail_in` would be `INT64_MAX` which is incorrect.

nmoinvaz:
> No problem. Just pls tell me how you'd prefer me to commit it: I guess not with --author="CodeRabbit", rather mentioning it in a Suggested-by: (Assisted-by:?)?

I have no preference. Mention the rabbit or not..

mz_strm_ppmd.c
solbjorn: I've been writing in C for 13 years and that's the first time I hear that `?` means `>= 0`.

https://godbolt.org/z/4nYEeaYs1

`?` always means `!= 0`.

nmoinvaz: Haha. I may have copied it wrong and confused it with JavaScript which I write in a lot these days.

solbjorn: Still, other decompressors consider `max_avail_in == 0` not as "nothing is available" but as "there's no limit". I think it's a rare case if possible at all since mz usually sets the limit via `MZ_STREAM_PROP_TOTAL_IN_MAX`.
So if we treat `max_avail_in == 0` as "nothing is available", it will be inconsistent with other decompression methods.

https://github.com/zlib-ng/minizip-ng/blob/develop/mz_strm_lzma.c#L172
https://github.com/zlib-ng/minizip-ng/blob/develop/mz_strm_zlib.c#L141
https://github.com/zlib-ng/minizip-ng/blob/develop/mz_strm_zstd.c#L113

nmoinvaz: There might be an issue with those too, so lets just leave it as you have it for now.

nmoinvaz (approved):

nmoinvaz:
I might clean up some of the ppmd code that I was looking through after this PR, this separate `reader()` function looks weird.

solbjorn:
> I might clean up some of the ppmd code that I was looking through after this PR, this separate `reader()` function looks weird.

It's a callback to be passed to the LZMA SDK code IIRC, the SDK is in general so messy so that I haven't even tried to add ZIP/7z/Rar support to my engine until I discovered minizip-ng and libarchive. No surprises there are issues when we try to reuse some parts of it.

Unfortunately, the author of it, Igor, is not community-friendly and publishes the 7-Zip code just for the sake of it (or to comply with [L]GPL). And vice versa, there have been forks adding Zstd support to 7-Zip for years but he just doesn't care and the latest release (26.00) still doesn't support it.

nmoinvaz:
I see ok.

Issue #914, opened 2025-10-29
[macOS] absolute path to MacOSX14.sdk is injected which fails downstream build with CMake


kambala-decapitator:
I've installed minizip-ng v4.0.10 from Homebrew and tried to use it in a CMake project. Configure step succeeds, but when I try to build something (used Ninja), I get:

> ninja: error: '/Library/Developer/CommandLineTools/SDKs/MacOSX14.sdk/usr/lib/libz.tbd', needed by 'bin/libvcmi.dylib', missing and no known rule to make it

CMake trace output for configure step revealed this:

/usr/local/lib/cmake/minizip-ng/minizip-ng.cmake(61):  set_target_properties(MINIZIP::minizip-ng PROPERTIES INTERFACE_INCLUDE_DIRECTORIES ${_IMPORT_PREFIX}/include/minizip-ng;${_IMPORT_PREFIX}/include/minizip-ng INTERFACE_LINK_DIRECTORIES /usr/local/Cellar/xz/5.8.1/lib;/usr/local/opt/zstd/lib INTERFACE_LINK_LIBRARIES /Library/Developer/CommandLineTools/SDKs/MacOSX14.sdk/usr/lib/libz.tbd;/Library/Developer/CommandLineTools/SDKs/MacOSX14.sdk/usr/lib/libbz2.tbd;lzma;zstd;/Library/Developer/CommandLineTools/SDKs/MacOSX14.sdk/System/Library/Frameworks/CoreFoundation.framework;/Library/Developer/CommandLineTools/SDKs/MacOSX14.sdk/System/Library/Frameworks/Security.framework;/Library/Developer/CommandLineTools/SDKs/MacOSX14.sdk/usr/lib/libiconv.tbd )

Coeur:
The `MacOSX` string does not exist anywhere in this project. So that sounds more like you facing an issue with cmake or ninja than an issue with minizip.

For better debugging, please tell us:
- What's your current macOS version? (command is `sw_vers`)
- What's your current cmake version? (command is `cmake --version`)
- What's your current ninja version? (command is `ninja --version`)

At the time of this post, I hope you have `cmake version 4.1.2` and ninja `1.13.1`, otherwise please upgrade (`brew upgrade` if you installed them via brew).

The minizip build instructions are:
https://github.com/zlib-ng/minizip-ng/blob/1867916bec205bb16299b7abe7145539eef6fb8a/README.md?plain=1#L62-L63

[...]

PR #983, opened 2026-04-25
cmake: produce portable installed package config

nmoinvaz:
Refactor the CMake build so the installed minizip-targets.cmake no longer bakes build-machine paths into INTERFACE_LINK_LIBRARIES. Each backend (zlib, zlib-ng, BZip2, LibLZMA, zstd, OpenSSL, Iconv, libbsd) now links via its imported target rather than raw \${FOO_LIBRARIES} / \${FOO_LIBRARY_DIRS} lists, the macOS CoreFoundation and Security frameworks are passed as -framework linker flags instead of SDK-absolute paths, and target_link_libraries is PRIVATE so backend deps stay scoped to the library. The generated config now only emits find_dependency() calls for packages located via find_package — fetched source trees are propagated through install(EXPORT) and don't need a consumer-side find_dependency. Fixes the breakage reported in #914 where Homebrew-installed minizip-ng pinned consumers to the build host's macOS SDK and Cellar paths.


VVD:
This code looks wrong: https://github.com/zlib-ng/minizip-ng/blob/develop/CMakeLists.txt#L404-L408
After this commit https://github.com/zlib-ng/minizip-ng/commit/a3c6e8c02dfefad636a58e43fdbb6caa4ddf49a9 `minizip-ng` linked with static library `libzstd.a`.
If replace this if with reversed:
```
        if(TARGET zstd::libzstd_shared)
            list(APPEND MINIZIP_LIB zstd::libzstd_shared)
        else()
            list(APPEND MINIZIP_LIB zstd::libzstd_static)
        endif()
```
then it linked dynamically with shared library `libzstd.so`.
Is `TARGET zstd::libzstd_{static|shared}` always true?

nmoinvaz:
Should be fixed in https://github.com/zlib-ng/minizip-ng/pull/989

VVD:
Thanks!

PR #992, opened 2026-05-01
update to use ppmd from 7zip 26.01

pmqs:
refresh with latest 7zip upstream source

### DIFF
--- a/.github/workflows/ppmd-dependabot.yml
+++ b/.github/workflows/ppmd-dependabot.yml
@@ -17,7 +17,8 @@ jobs:
     - name: Check Upstream PPMd Version
       run: |
         # This is the hard-wired PPMd version that minizip-ng is built with.
-        expected_ppmd_version="26.00"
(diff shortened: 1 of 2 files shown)


nmoinvaz:
Does the dependabot script open PRs with the new versions?

pmqs:
> Does the dependabot script open PRs with the new versions?

Afraid not. That would need cmake support from dependabot and it isn't there at the moment.

There is an open ticket to get cmake support added , but that is now a few years old. See https://github.com/dependabot/dependabot-core/issues/7451

All this quick-and-dirty dependabot does is fail if it finds the hard-wired version is different upstream.

I only spotted this because I got a failure email from my fork of this repo.

Think we decided to keep with explicit versions of the 7zip source because there was a breaking API change at one point with the PPMd source. 

The alternative is to clone the latest code from upstream and deal with any failures as they happen.

nmoinvaz:
I think it might not be that hard to have the current workflow create a PR. I can take a look at it later I guess.

nmoinvaz (approved):

PR #1011, opened 2026-07-02
cmake: option MZ_PBKDF2: export mz_crypt_pbkdf2

sergio-nsk:
Let users use the function if they implement `mz_crypt_hmac_*` API without using CommonCrypto or OpenSSL.

### DIFF
--- a/mz_crypt.c
+++ b/mz_crypt.c
@@ -91,7 +91,7 @@ uint32_t mz_crypt_crc32_update(uint32_t value, const uint8_t *buf, int32_t size)
 #endif
 }
 
-#if defined(HAVE_WZAES)
+#if defined(HAVE_PBKDF2)
 int32_t mz_crypt_pbkdf2(const uint8_t *password, int32_t password_length, const uint8_t *salt, int32_t salt_length,
                         uint32_t iteration_count, uint8_t *key, uint16_t key_length) {
     void *hmac1 = NULL;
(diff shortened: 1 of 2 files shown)


sergio-nsk:
@coderabbitai Please update your "Walkthrough" comment,

mz_crypt.c
```diff
@@ -91,7 +91,7 @@ uint32_t mz_crypt_crc32_update(uint32_t value, const uint8_t *buf, int32_t size)
 #endif
 }
 
-#if defined(HAVE_WZAES)
+#if defined(HAVE_PBKDF2)
```
nmoinvaz, line 94: Let's remove all CMake changes. Then do `#if !define(NO_PBKDF2)`. 

Then in our CMake, let's use `set_target_properties` on minizip-ng to set `-D NO_PBKDF2`.

I don't think this is worthy of a feature flag.

nmoinvaz (changes requested):

mz_crypt.c
sergio-nsk: I guess you have meant a macro and a condition like this:
```c
// Don't remove the right hand of the condition. Client apps may use the guarded function.
#if defined(HAVE_WZAES) || defined(HAVE_PBKDF2)
```

sergio-nsk:
I can `-DHAVE_WZAES` on the source file in the minizip-ng subdirectory.

PR #1014, opened 2026-07-03
Allow recover flag to read zips with >65535 entries missing zip64 EOCD

### LINKED ISSUE #597: Read corrupted zip files with more than 65535 entries
Old minizip (1.1 with zip64 patch) was able to generate zip files that have more than 65535 entries without switching to zip64 mode. So result zip doesn't have ZIP64 central directory. At the same time it wasn't giving any errors and was able to fully read it afterwards. I understand that this is considered to be improper zip, but is it possible to read such files with new minizip-ng?
I had to omit format sanity check to read all entries without errors (replaced with err = MZ_OK;
https://github.com/zlib-ng/minizip-ng/blob/99d39015e29703af2612277180ea586805f655ea/mz_zip.c#L1038

Is there some better way? Can it be added as option? I also tried using "recover" flag but it was still giving error.
Here is an example of such file (65536 entries of empty files):
[corrupted_65536_entries.zip](https://github.com/zlib-ng/minizip-ng/files/7392713/corrupted_65536_entries.zip)

tbeu:
Resolves #597 

When number_entry in the standard EOCD is UINT16_MAX, the reader looks for a zip64 EOCD record. If not found, it returned MZ_FORMAT_ERROR even when the recover flag was set. Additionally, recovery scanning started from an EOF-adjacent position after the failed zip64 search, missing local headers.

Fix: respect the recover flag in the zip64-not-found error path, and seek to the start of the file before recovery scanning.

### DIFF
--- a/mz_zip.c
+++ b/mz_zip.c
@@ -1057,7 +1057,8 @@ static int32_t mz_zip_read_cd(void *handle) {
                 }
             } else if ((zip->number_entry == UINT16_MAX) || (number_entry_cd != zip->number_entry) ||
                        (zip->cd_size == UINT16_MAX) || (zip->cd_offset == UINT32_MAX)) {
-                err = MZ_FORMAT_ERROR;
+                if (!zip->recover)
+                    err = MZ_FORMAT_ERROR;
             }
         }
     }
@@ -1267,11 +1268,13 @@ static int32_t mz_zip_recover_cd(void *handle) {
 
     /* Determine if we are on a split disk or not */
     mz_stream_set_prop_int64(zip->stream, MZ_STREAM_PROP_DISK_NUMBER, 0);
-    if (mz_stream_tell(zip->stream) < 0) {
+    if (mz_stream_tell(zip->stream) < 0)
         mz_stream_set_prop_int64(zip->stream, MZ_STREAM_PROP_DISK_NUMBER, -1);
-        mz_stream_seek(zip->stream, 0, MZ_SEEK_SET);
-    } else
+    else
         disk_number_with_cd = 1;
+
+    /* Always seek to start of stream for recovery scanning */
+    mz_stream_seek(zip->stream, 0, MZ_SEEK_SET);
 
     local_file_info_stream = mz_stream_mem_create();
     if (!local_file_info_stream)


mz_zip.c
```diff
@@ -1452,8 +1453,11 @@ int32_t mz_zip_open(void *handle, void *stream, int32_t mode) {
             err = mz_zip_read_cd(zip);
             if (err != MZ_OK) {
                 mz_zip_print("Zip - Error detected reading cd (%" PRId32 ")\n", err);
-                if (zip->recover && mz_zip_recover_cd(zip) == MZ_OK)
-                    err = MZ_OK;
+                if (zip->recover) {
+                    mz_stream_seek(zip->stream, 0, MZ_SEEK_SET);
+                    if (mz_zip_recover_cd(zip) == MZ_OK)
```
nmoinvaz: `mz_zip_recover_cd` does call `mz_stream_seek(zip->stream, 0, MZ_SEEK_SET);` so maybe that logic needs to go inside `mz_zip_recover_cd`?

mz_zip.c
```diff
@@ -1055,8 +1055,9 @@ static int32_t mz_zip_read_cd(void *handle) {
                     if (zip->cd_offset < 0)
                         err = MZ_FORMAT_ERROR;
                 }
-            } else if ((zip->number_entry == UINT16_MAX) || (number_entry_cd != zip->number_entry) ||
-                       (zip->cd_size == UINT16_MAX) || (zip->cd_offset == UINT32_MAX)) {
+            } else if (!zip->recover &&
```
nmoinvaz: let's leave the if statement as-is. But add `if (!zip->recover) err = MZ_FORMAT_ERROR` inside of it. I think it will be easier to read to split them up into two if statements.

mz_zip.c
```diff
@@ -1265,6 +1266,9 @@ static int32_t mz_zip_recover_cd(void *handle) {
     if (err != MZ_OK)
         return err;
 
+    /* Always seek to start of stream for recovery scanning */
+    mz_stream_seek(zip->stream, 0, MZ_SEEK_SET);
```
nmoinvaz: This only needs to happen after `disk_number_with_cd = 1`, right, because it is already called on line 1276. Please squash all commits.

tbeu:
Reproducer

```c
#include "mz.h"
#include "mz_strm.h"
#include "mz_strm_os.h"
#include "mz_zip.h"
#include "mz_zip_rw.h"
#include <stdio.h>
int main(void) {
    void *s = mz_stream_os_create();
    void *r = mz_zip_reader_create();
    mz_zip_reader_set_recover(r, 1);
    int32_t e = mz_stream_os_open(s, "corrupted_65536_entries.zip", MZ_OPEN_MODE_READ);
    e = mz_zip_reader_open(r, s);
    printf("%s (err=%d)\n", e == MZ_OK ? "OPEN OK" : "OPEN FAIL", e);
    if (e) return 1;
    e = mz_zip_reader_goto_first_entry(r);
    int64_t n = 0;
    while (e == MZ_OK) { n++; e = mz_zip_reader_goto_next_entry(r); }
    printf("entries=%" PRId64 " %s\n", n, n == 65536 ? "PASS" : "FAIL");
    return n != 65536;
}
```

mz_zip.c
```diff
@@ -1269,10 +1270,12 @@ static int32_t mz_zip_recover_cd(void *handle) {
     mz_stream_set_prop_int64(zip->stream, MZ_STREAM_PROP_DISK_NUMBER, 0);
     if (mz_stream_tell(zip->stream) < 0) {
```
nmoinvaz: no need for braces here, sorry :-(

PR #1016, opened 2026-07-04
Add MZ_EXPORT to public API symbols and hide internal symbols

### LINKED ISSUE #475: Windows: DLL export mz functions in mz_zip.h

tbeu:
- Update MZ_EXPORT macro to use visibility attributes on non-Windows
- Add MZ_EXPORT to all public function declarations in headers
- Set C_VISIBILITY_PRESET hidden on minizip target
- Fix ZEXPORT in compat layer to properly export symbols
- Add missing ZEXPORT on compat API functions

Resolves #475

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -715,7 +715,8 @@ set_target_properties(${MINIZIP_TARGET} PROPERTIES
                       VERSION ${VERSION}
                       SOVERSION ${SOVERSION}
                       LINKER_LANGUAGE C
-                      DEFINE_SYMBOL "MZ_EXPORTS")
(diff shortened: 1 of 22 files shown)


nmoinvaz:
I also don't like this change putting MZ_EXPORT in so many places. I think the original issue has the correct solution for CMake.

tbeu:
Well, it does not hurt. For me this is another convenience API for any consumer. I'd consider it useful. But it's up to you.

PR #1017, opened 2026-07-04
Support custom memory allocation functions like zlib-ng

### LINKED ISSUE #794: Support custom memory allocation functions
At the moment, *_create functions use `calloc`. Please support custom memory (de)allocation functions like zlib-ng does.

tbeu:
Add mz_alloc_func, mz_free_func, mz_realloc_func, mz_strdup_func typedefs and per-instance setters (mz_stream_set_alloc_funcs, mz_zip_set_alloc_funcs, mz_zip_reader_set_alloc_funcs, mz_zip_writer_set_alloc_funcs) that allow applications to override the C standard library allocators. All internal calloc/malloc/free/realloc/strdup calls now route through MZ_ALLOC/MZ_FREE/MZ_REALLOC/MZ_STRDUP macros, falling back to stdlib defaults when no custom allocator is set.
Custom allocators are also propagated to internal zlib-ngstreams.

Resolves #794

### DIFF
--- a/compat/ioapi.c
+++ b/compat/ioapi.c
@@ -200,7 +200,7 @@ int32_t mz_stream_ioapi_set_filefunc64(void *stream, zlib_filefunc64_def *filefu
 }
 
 void *mz_stream_ioapi_create(void) {
-    mz_stream_ioapi *ioapi = (mz_stream_ioapi *)calloc(1, sizeof(mz_stream_ioapi));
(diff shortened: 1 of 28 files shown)


nmoinvaz:
I'm not sure I like the complexity. I am considering just closing the feature request as won't support or I can point them to this PR.

tbeu:
Hm, I consider it a convenience layer. zlib-ng has it as well. However I have to admit that I also do not like the malloc/calloc issues. Ideally there should only be one callback.

nmoinvaz:
I have decided not to do this. I will close the GH issue as won't support.

PR #1018, opened 2026-07-04
Add MZ_ICU option for string encoding conversion

### LINKED ISSUE #998: Option to use ICU instead of libiconv for string encoding conversion
Currently, minizip-ng relies on libiconv for character set conversions. It would be useful to have an option to use ICU (International Components for Unicode) as an alternative backend. ICU is widely adopted, actively maintained, and provides robust Unicode handling across platforms.

tbeu:
Add ICU as an alternative backend to libiconv using ucnv_convert. ICU is detected via CMake's FindICU module on non-Windows platforms. Add CI build configuration for ICU in GitHub Actions.

Resolves #998

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -60,6 +60,7 @@ option(MZ_OPENSSL "Enables OpenSSL for encryption" ${UNIX})
 cmake_dependent_option(MZ_LIBBSD "Builds with libbsd crypto random" ON "UNIX" OFF)
 # Character conversion options
 option(MZ_ICONV "Enables iconv for string encoding conversion" ON)
+option(MZ_ICU "Enables ICU for string encoding conversion" OFF)
 # Code generation options
 option(MZ_COMPRESS_ONLY "Only support compression" OFF)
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -60,6 +60,7 @@ option(MZ_OPENSSL "Enables OpenSSL for encryption" ${UNIX})
 cmake_dependent_option(MZ_LIBBSD "Builds with libbsd crypto random" ON "UNIX" OFF)
 # Character conversion options
 option(MZ_ICONV "Enables iconv for string encoding conversion" ON)
+option(MZ_ICU "Enables ICU for string encoding conversion" OFF)
```
pmqs, line 63: What happens if both MZ_ICONV and MZ-ICU are set to ON?

tbeu: Will add mutual exclusion.

nmoinvaz (approved):

nmoinvaz:
Squash commits please.

tbeu:
I believe the new CI check will only run once merged. Hence, there still is one unknown here.

nmoinvaz:
Thanks!

PR #1022, opened 2026-07-15
Harden extraction against destination symlink traversal

Alearner12:
# Harden archive extraction against destination symlink traversal

## Summary

This change makes archive extraction consistently enforce the destination-root containment policy for directories, regular files, and archive-created symbolic links.

Previously, ordinary directory entries returned through recursive directory creation before the existing symlink-containment check. On POSIX systems, an existing destination symlink could therefore redirect directory creation outside the extraction root. Regular-file extraction also opened paths without no-follow semantics, and the existing target check was lexical rather than filesystem-aware for nested symlink chains.

## Security and correctness impact

The affected pattern is:

```text
validate a pathname
    -> later traverse or reopen it through the filesystem
```

The patch enforces the following invariants during extraction:

1. Every extraction entry checks existing parent components before mutation.
2. Existing nested symlink targets are resolved through the filesystem where supported and must remain inside the destination root.
3. Regular-file output cannot use an existing final symlink.
4. POSIX directory and file operations traverse parents through descriptor-relative no-follow operations.
5. POSIX archive-created symlinks are created relative to a safely opened parent directory.
6. Windows extraction rejects existing reparse points for no-follow file opens and performs the corresponding path checks.

## Relevant implementation changes

- `mz_zip_rw.c`
  - Runs `mz_dir_has_unsafe_symlink()` before the directory-entry early return.
  - Uses `mz_os_make_dir_safe()` for extraction directory creation.
  - Opens extracted regular files with `MZ_OPEN_MODE_NOFOLLOW`.
  - Uses `mz_os_make_symlink_safe()` for archive-created links.

- `mz_os.c`
  - Resolves existing symlink targets through `mz_os_get_real_path()` before comparing them with the canonical destination root.

- `mz_os_posix.c` and `mz_strm_os_posix.c`
  - Add descriptor-relative `openat()`/`mkdirat()`/`symlinkat()` traversal with `O_NOFOLLOW`.

- `mz_os_win32.c` and `mz_strm_os_win32.c`
  - Resolve existing paths through Windows file handles where supported.
  - Reject existing reparse points when no-follow extraction is requested.

- Tests
  - Reject nested existing symlink escapes.
  - Reject existing symlink components during extraction checks.
  - Reject final symlinks on safe file open.

## Deterministic reproducer addressed

Given:

```text
dest/redirect -> outside
archive entry: link -> redirect
archive entry: link/victim
```

The previous implementation accepted `dest/link`, then opened `dest/link/victim` with ordinary `open()`, allowing the write below `outside`. The patched implementation rejects the symlink target before creating `link` or writing `victim`.

For a directory entry:

```text
dest/pivot -> outside
archive entry: pivot/newdir/
```

the patched implementation rejects the entry before `mkdirat()` can create `outside/newdir`.

In-root symlinks remain supported. For example, an archive link pointing to another path within the extraction root is still created successfully.

## Verification

Executed on the POSIX build in this workspace:

- Full baseline CTest suite: **69/69 passed**.
- Focused GoogleTest suite under ASan (`ASAN_OPTIONS=detect_leaks=0`): **passed**.
- Focused GoogleTest suite under UBSan: **passed**.
- Manual directory-symlink escape reproducer: rejected; no outside directory created.
- Manual nested-symlink reproducer: rejected; no outside file created.
- Manual in-root symlink extraction: passed.

## Scope and limitations

This PR establishes a consistent extraction containment policy and closes the demonstrated POSIX traversal paths. Windows directory creation and post-write timestamp/attribute updates still use platform pathname APIs; complete cross-platform elimination of every possible concurrent pathname race would require additional handle-based metadata APIs.

The change intentionally preserves legitimate in-root symbolic-link extraction and does not alter archive creation behavior.

### DIFF
--- a/test/test_path.cc
+++ b/test/test_path.cc
@@ -164,4 +164,13 @@ TEST(path_security, rejects_existing_symlink_component_escape) {
     rmdir(base.c_str());
     rmdir(root);
 }
+
+TEST(path_security, accepts_filesystem_root_as_containment_base) {
+    EXPECT_EQ(mz_path_is_symlink_target_safe("/link", "tmp/target", "/"), MZ_OK);
+}
+
+TEST(path_security, rejects_overlong_symlink_path) {
+    std::string long_path(1100, 'a');
+    EXPECT_EQ(mz_path_is_symlink_target_safe(long_path.c_str(), "target", "."), MZ_BUF_ERROR);
+}
 #endif
--- a/test/test_path.cc
+++ b/test/test_path.cc
@@ -15,6 +15,7 @@
 #include "mz_zip_rw.h"
 
 #include <algorithm>
+#include <cstdlib>
 #include <cstring>
 #include <gtest/gtest.h>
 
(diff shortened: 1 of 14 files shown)


nmoinvaz:
Have you confirmed the issue empirically? There is too much changes.

Alearner12:
@nmoinvaz   Yes, I confirmed the issue empirically on POSIX against develop. With dest/pivot as a symlink to an outside directory and a ZIP containing the directory entry pivot/newdir/, mz_zip_reader_save_all() returns success and creates outside/newdir. Directory entries return through mz_dir_make() before reaching the existing mz_dir_has_unsafe_symlink() check. I agree the current patch is too broad. I will reduce the PR to moving the existing check before the directory-entry branch and add a deterministic regression test. I will leave the broader no-follow/TOCTOU redesign for a separate discussion.

nmoinvaz:
All these symlink checks are too messy and I think it might be better to just remove that feature altogether.

Alearner12:
Understood. To clarify, do you mean removing support for extracting symbolic-link entries entirely, or keeping symbolic-link entries but rejecting any extraction path that traverses an existing symlink? I confirmed the directory-entry behavior empirically: with dest/pivot -> outside and archive entry pivot/newdir/, extraction succeeds and creates outside/newdir.

Alearner12:
@nmoinvaz Following up on this removing archive-created symlinks alone would not fix the demonstrated directory-entry escape, because dest/pivot -> outside already exists before extraction.My proposed interpretation is to reject symbolic-link archive entries, reject extraction through any existing symlink component, and remove the complicated “safe in-root symlink” target-containment logic.This would make extraction fail closed and substantially reduce the code.  Is that the direction you intended?

nmoinvaz:
The code is too hard to read.

Alearner12:
@nmoinvaz Thanks , I've reduced the pr to minimal fix 

- moved the existing mz_dir_has_unsafe_symlink() check before the directory-entry early return;
-  added one deterministic regression test for destination/pivot -> outside with archive entry pivot/newdir/;
-  removed all new path-resolution, no-follow, and platform-specific APIs.

Alearner12:
@nmoinvaz , gentle ping on this pr , when u have time

test/test_path.cc
```diff
@@ -66,6 +76,98 @@ TEST_P(path_resolve, os) {
     EXPECT_STREQ(output, expected_path.c_str());
 }
 
+#if !defined(_WIN32) && defined(HAVE_SYMLINK)
+class zip_reader_symlink_test : public ::testing::Test {
```
nmoinvaz: `test_path.cc` is really for `mz_path*` functions. I think we need to move this test into something like `test_reader.cc`...

nmoinvaz (commented):
the actual symlink fix looks good. just need to move the unit tests to a different file.

Alearner12:
@nmoinvaz Done! Moved the extraction symlink unit tests into `test/test_reader.cc` and updated `test/CMakeLists.txt`.

Alearner12:
I noticed several related extraction-security issues are now linked to this PR, and some of them overlap with cases the earlier broader version of this patch was trying to address.
Since this PR is now intentionally kept minimal, would you prefer that I work on those remaining cases as separate focused PRs, each with its own reproducer and regression test? If multiple issues turn out to share the same root cause, I can also propose a small common abstraction after the individual cases are understood. I just want to avoid expanding the scope here again without ur  guidance.

nmoinvaz:
TBH these types of PRs are annoying and makes me want to remove symlinks or remove features or just shutdown this project. I haven't decided yet. This project was supposed to be simple but now it has gotten out of hand.

nmoinvaz:
I think maybe CMake needs a feature flag for symlinks so it can be turned off.

nmoinvaz:
Looks like lint is failing. After it is fixed I can merge.

Alearner12:
> Looks like lint is failing. After it is fixed I can merge.

was formatting issue into two test files , fixed

nmoinvaz:
Thanks for this PR and pushing it over the finish line. Feel free to address any of the remaining issues in small PRs.

Issue #154, opened 2017-09-29
This zip file cannot be unzip (GB2312 encoding)


Coeur:
https://github.com/ZipArchive/ZipArchive/files/769125/Calculator.zip

Results with other tools:  
<img width="346" alt="screen shot 2017-09-29 at 6 38 05 pm" src="https://user-images.githubusercontent.com/839992/31012914-44a75bba-a4d9-11e7-804b-fa86129fdf97.png">

Results with miniunz:  
>Calculator.zip opened
error -103 with zipfile in unzOpenCurrentFilePassword
error -102 with zipfile in unzCloseCurrentFile
error -103 with zipfile in unzGoToNextFile

So, miniunz is the only tool unable to unzip this file at all.

In [unzip.c, line 1019](https://github.com/nmoinvaz/minizip/blob/master/unzip.c#L1019), the test `size_filename != s->cur_file_info.size_filename` is failing, as size_filename from filestream ([set on line 1018](https://github.com/nmoinvaz/minizip/blob/master/unzip.c#L1017)) is 27 but from filestream_with_CD ([set on line 751](https://github.com/nmoinvaz/minizip/blob/master/unzip.c#L751)) it's 24.

Commenting out lines 1019 and 1020 and ignoring the differences in size_filename let the file be unzipped correctly.

nmoinvaz:
I’m not sure how important even checking for local header consistency is. It may be possible to get rid of the check some day.

PR #1024, opened 2026-07-22
Fix non-ASCII filename handling on Windows (for CJK)

serunis:
I discovered a lack of Korean character support while running this program on a Korean Windows setup, and I have applied a patch generated via Claude to resolve it. Because the Windows console does not default to the UTF-8 code page, it is highly likely that all non-ASCII users are encountering this same problem. I would appreciate your review on this PR. The patch details provided by Claude are as follows.

## Problem

There are three independent issues when working with non-ASCII filenames on
Windows (reproduced on Korean Windows with CP949, but the same applies to any
ANSI code page locale — CP932/936/950/1251/1252, etc.):

1. **CLI cannot open non-ASCII paths passed on the command line.**
   The CRT supplies `argv` to `main()` in the ANSI code page, while minizip-ng
   treats every `char *` string as UTF-8 (`MultiByteToWideChar(CP_UTF8, ...)`
   in `mz_os_win32.c`). The ANSI bytes form an invalid UTF-8 sequence, so the
   path is converted with U+FFFD replacement characters and `CreateFileW`
   fails with `MZ_OPEN_ERROR` (-111).

2. **Extracting zips created by legacy archivers permanently corrupts
   filenames.** Windows Explorer and many archivers store filenames in the
   system ANSI code page without setting the UTF-8 flag (general purpose
   bit 11). minizip-ng interprets those bytes as UTF-8 when no encoding is
   specified, producing mojibake filenames on disk (e.g. `�ѱ�����.txt`
   instead of `한글파일.txt`) that cannot be recovered afterwards.

3. **Console output displays UTF-8 filenames as mojibake** when the console
   code page is not UTF-8 (display-only issue).

## Changes

- **`mz_os_utf8_string_is_valid()`** (`mz_os.c`): strict portable UTF-8
  validator — rejects invalid lead/continuation bytes, overlong encodings,
  surrogates, and code points beyond U+10FFFF.
- **`mz_os_get_default_encoding()`**: returns the system default ANSI code
  page on Windows (`GetACP()`), `0` on POSIX platforms where the concept
  does not exist (fallback disabled, behavior unchanged).
- **Conservative encoding fallback** in `mz_zip_reader_save_all()` and the
  minizip CLI display paths. It only triggers when **all three** hold:
  1. no encoding was explicitly specified,
  2. the entry's UTF-8 flag is not set,
  3. the stored filename is not a valid UTF-8 sequence.

  ```c
  entry_encoding = reader->encoding;
  if ((entry_encoding <= 0) && (flag & MZ_ZIP_FLAG_UTF8) == 0 &&
      mz_os_utf8_string_is_valid(filename) != MZ_OK)
      entry_encoding = mz_os_get_default_encoding();
  ```

  An explicitly specified encoding always takes precedence, valid UTF-8
  names are never touched, and POSIX behavior is unchanged.
- **minizip CLI (Windows)**: re-fetch the command line with
  `GetCommandLineW()` + `CommandLineToArgvW()` and convert each argument to
  UTF-8 at startup, with both `WideCharToMultiByte` calls checked and all
  allocations freed on the failure path; switch the console output code page
  to UTF-8, saving the original code page and restoring it with an `atexit`
  handler so the change does not leak into the calling console session.
- **`MZ_ENCODING_CODEPAGE_949`** added to `mz.h` (932/936/950 were already
  defined; 949 was missing).

## Testing

Built with GCC (w64devkit) + Ninja on Windows 11. Test zips were created
directly with the library API: one UTF-8 zip with the UTF-8 flag set, and one
CP949 zip (raw bytes `C7 D1 B1 DB C6 C4 C0 CF` = "한글파일") without the flag.

| Scenario | Before | After |
|---|---|---|
| `minizip -o test.zip 한글파일.txt` | Error -111 | ✅ compresses, stores UTF-8 name + flag |
| Console output | mojibake | ✅ correct |
| Extract CP949 zip without `-c` | corrupted filename on disk | ✅ restored automatically |
| List CP949 zip (`-l`) | mojibake | ✅ correct |
| Extract UTF-8 zip (regression) | ✅ | ✅ unchanged (fallback blocked by flag check) |
| Explicit `-c 949` | ✅ | ✅ still takes precedence |
| ASCII filenames (regression) | ✅ | ✅ unchanged |
| Compress → extract roundtrip | — | ✅ names and contents match |
| Console code page after exit | leaked CP_UTF8 into the calling session | ✅ restored (verified from CP949 and CP65001 consoles) |

## Why this belongs upstream

- **This class of bug is invisible to ASCII-locale development and CI.**
  ASCII bytes are identical in every code page and are also valid UTF-8, so
  none of these failures can reproduce with ASCII filenames. Users who can
  observe the problem have reported it repeatedly over the years —
  #154 (GB2312 zip cannot be extracted), #328 (legacy CJK/DBCS code page
  support), #868 (support any character encoding), #1018 (ICU-based
  conversion attempt) — but it needs a fix from a locale that can test it.

- **Not Korean-specific.** The fallback uses `GetACP()`, so the same code
  restores Shift-JIS zips on Japanese Windows (CP932), GBK on Chinese
  Windows (CP936), Cyrillic on Russian Windows (CP1251), accented Latin
  names on Western European Windows (CP1252), and so on. The prior reports
  above were mostly Chinese; CP949 is only what this patch was verified on.

- **It aligns minizip-ng with de facto ecosystem behavior.** Windows
  Explorer, 7-Zip, and Info-ZIP unzip all interpret entry names without the
  UTF-8 flag in the local code page. minizip-ng is currently the outlier
  when reading such archives, and unlike display bugs this one is
  destructive: the mojibake names are written to disk.

- **Conservative by construction, no behavior change for existing users.**
  The fallback triggers only when no encoding was specified *and* the UTF-8
  flag is absent *and* the name fails strict UTF-8 validation; an explicit
  `-c` always wins, valid UTF-8 names are never touched, POSIX behavior is
  unchanged, and there is no new dependency, API, or ABI change (one added
  helper pair in `mz_os`). The library's "all `char *` strings are UTF-8"
  contract is preserved — the CLI argv conversion *restores* that contract
  at the process boundary rather than weakening it.

## Known limitations

- If a legacy code page byte sequence happens to also be valid UTF-8, it is
  treated as UTF-8. The zip format carries no encoding metadata, so perfect
  detection is impossible; `-c <codepage>` can be used to force a conversion
  as before. This coincidence is rare for multi-byte CJK code pages.
- The fallback uses the *system* code page, so e.g. extracting a CP949 zip on
  Japanese Windows attempts CP932. `-c 949` handles that case.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01EDxibo7seKgSHceGNoHqHB


serunis:
I forgot to restore console code page, I will update

mz_os.c
```diff
@@ -492,4 +492,51 @@ int32_t mz_file_get_crc(const char *path, uint32_t *result_crc) {
...
+
+    if (!string)
+        return MZ_PARAM_ERROR;
+
+    while (*s != 0) {
+        c = *s++;
+
+        if (c < 0x80) {
+            continue;
+        } else if ((c & 0xe0) == 0xc0) {
```
nmoinvaz: remove `else if` and braces around continue;

mz_os.h
```diff
@@ -127,6 +130,10 @@ char *mz_os_utf8_string_create(const char *string, int32_t encoding);
 void mz_os_utf8_string_delete(char **string);
 /* Delete a utf8 string that was created */
 
+int32_t mz_os_get_default_encoding(void);
+/* Gets the system default (ANSI) code page for legacy string conversion,
+   returns 0 if not available */
```
nmoinvaz: no need to mention return value, no other comments do.

mz_os.h
```diff
@@ -112,6 +112,9 @@ int32_t mz_dir_make(const char *path);
 int32_t mz_file_get_crc(const char *path, uint32_t *result_crc);
 /* Gets the crc32 hash of a file */
 
+int32_t mz_os_utf8_string_is_valid(const char *string);
```
nmoinvaz, line 115: We need to add these two new functions to the docs dir.

mz_os.h
```diff
@@ -112,6 +112,9 @@ int32_t mz_dir_make(const char *path);
...
+/* Returns MZ_OK if the string is a valid utf8 byte sequence */
```
nmoinvaz: Replace "Returns MZ_OK" -> "Checks"

mz_zip_rw.c
```diff
@@ -904,8 +905,16 @@ int32_t mz_zip_reader_save_all(void *handle, const char *destination_dir) {
         strncpy(utf8_name, reader->file_info->filename, utf8_name_size - 1);
         utf8_name[utf8_name_size - 1] = 0;
 
-        if ((reader->encoding > 0) && (reader->file_info->flag & MZ_ZIP_FLAG_UTF8) == 0) {
-            utf8_string = mz_os_utf8_string_create(reader->file_info->filename, reader->encoding);
+        entry_encoding = reader->encoding;
+        if ((entry_encoding <= 0) && (reader->file_info->flag & MZ_ZIP_FLAG_UTF8) == 0 &&
+            mz_os_utf8_string_is_valid(reader->file_info->filename) != MZ_OK) {
+            /* Not valid utf8 and no encoding specified; fall back to the
```
nmoinvaz: fallback one word. Claude loves 2 words

mz_zip_rw.c
```diff
@@ -904,8 +905,16 @@ int32_t mz_zip_reader_save_all(void *handle, const char *destination_dir) {
         strncpy(utf8_name, reader->file_info->filename, utf8_name_size - 1);
         utf8_name[utf8_name_size - 1] = 0;
 
-        if ((reader->encoding > 0) && (reader->file_info->flag & MZ_ZIP_FLAG_UTF8) == 0) {
-            utf8_string = mz_os_utf8_string_create(reader->file_info->filename, reader->encoding);
+        entry_encoding = reader->encoding;
+        if ((entry_encoding <= 0) && (reader->file_info->flag & MZ_ZIP_FLAG_UTF8) == 0 &&
+            mz_os_utf8_string_is_valid(reader->file_info->filename) != MZ_OK) {
```
nmoinvaz: Do we really need to check the entire string??

minizip.c
```diff
@@ -153,8 +159,16 @@ int32_t minizip_list(const char *path, int32_t encoding) {
         method = mz_zip_get_compression_method_string(file_info->compression_method);
         mz_zip_time_t_to_tm(file_info->modified_date, &tmu_date);
 
-        if ((encoding > 0) && (file_info->flag & MZ_ZIP_FLAG_UTF8) == 0) {
-            utf8_string = mz_os_utf8_string_create(file_info->filename, encoding);
+        entry_encoding = encoding;
+        if ((entry_encoding <= 0) && (file_info->flag & MZ_ZIP_FLAG_UTF8) == 0 &&
+            mz_os_utf8_string_is_valid(file_info->filename) != MZ_OK) {
```
nmoinvaz, line 164: Do we need to check the entire string here?

minizip.c
```diff
@@ -588,6 +648,18 @@ int main(int argc, const char *argv[]) {
     const char *destination = NULL;
     const char *filename_to_extract = NULL;
 
+#if defined(_WIN32)
+    /* The console code page is usually not UTF-8 so UTF-8 filenames
+       would display incorrectly */
+    SetConsoleOutputCP(CP_UTF8);
+
+    argv = minizip_utf8_argv_create(&argc);
```
nmoinvaz: I wonder if it makes sense to use `WinMainW`.

nmoinvaz (changes requested):
Good start. I think the last 2 comments might be amendable to previous commits in the series.

minizip.c
serunis: I guess you mean wmain, I am not sure. I can try if it works and simplify the code

minizip.c
serunis: Yes. A CP949 filename such as final_report_realfinal_진짜최종.docx has a valid ASCII prefix but invalid UTF-8 bytes later in the string. We need to inspect the whole filename to determine whether it is valid UTF-8.

mz_os.c
serunis: done

mz_os.h
serunis: done

mz_os.h
serunis: add description to mz_os.md

mz_os.h
serunis: done

mz_zip_rw.c
serunis: Done. The wording was removed when the fallback logic was consolidated into mz_zip_reader_get_entry_encoding().

minizip.c
serunis: I submit the code

minizip.c
```diff
@@ -575,7 +587,55 @@ int32_t minizip_erase(const char *src_path, const char *target_path, int32_t arg
...
+/* Converts the Unicode command line arguments supplied to wmain to UTF-8. */
+static const char **minizip_utf8_argv_create(int32_t argc, wchar_t *argv_wide[]) {
+    const char **argv_utf8 = NULL;
+    int32_t arg_utf8_size = 0;
+    int32_t i = 0;
+
+    argv_utf8 = (const char **)calloc(argc + 1, sizeof(char *));
+    if (argv_utf8) {
+        for (i = 0; i < argc; i += 1) {
+            arg_utf8_size = WideCharToMultiByte(CP_UTF8, 0, argv_wide[i], -1, NULL, 0, NULL, NULL);
```
nmoinvaz: this can be `mz_os_utf8_string_create_from_unicode` right?

mz_os.c
```diff
@@ -492,4 +492,54 @@ int32_t mz_file_get_crc(const char *path, uint32_t *result_crc) {
     return err;
 }
 
+/* Checks that a string is a valid utf8 byte sequence; rejects invalid lead and
+   continuation bytes, overlong encodings, surrogates, and code points beyond
+   U+10FFFF. Returns MZ_OK if valid. */
```
nmoinvaz: Remove: "Returns MZ_OK if valid."

minizip.c
nmoinvaz: Ok that makes sense.

nmoinvaz (changes requested):
This is looking good, just need a few more adjustments.

nmoinvaz:
Looks like the Linting CI failed. All other changes you made look good!

nmoinvaz (approved):

nmoinvaz:
Thanks!

PR #1034, opened 2026-09-01
Fix compression test error propagation

serunis:
## Summary

- Stop archive creation at the first input failure so the CLI preserves and returns the original error.
- Update the compression CTest fixtures to use files that exist and erase entries that were actually archived.
- Skip compression test generation when no compression methods are available.

## Details

The compression tests referenced `test.c` and `test.h`, which do not exist in the test working directory. `minizip_add()` reported those failures but continued processing later inputs. A later successful addition could overwrite the earlier error, causing the command to return success after a partial failure and masking the invalid test fixtures.

This change makes `minizip_add()` stop at the first failed path so callers and CTest receive the original non-zero result. Archive close failures are still reported, but no longer replace an earlier input error. The test inputs now use existing fixtures, and the erase tests remove entries that were actually added to the archive.

CTest generation also returns early for variants that have no available compression method, instead of creating duplicate `NOTFOUND-*` tests.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -919,6 +919,9 @@ if(MZ_BUILD_TESTS AND MZ_BUILD_UNIT_TESTS)
             list(APPEND COMPRESS_METHOD_ARGS "-t")
         endif()
         list(LENGTH COMPRESS_METHOD_NAMES COMPRESS_METHOD_COUNT)
+        if(COMPRESS_METHOD_COUNT EQUAL 0)
+            return()
+        endif()
         math(EXPR COMPRESS_METHOD_COUNT "${COMPRESS_METHOD_COUNT}-1")
         foreach(INDEX RANGE ${COMPRESS_METHOD_COUNT})
             list(GET COMPRESS_METHOD_NAMES ${INDEX} COMPRESS_METHOD_NAME)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -95,8 +95,11 @@ cmake --build build
 
 ## Third-Party Libraries
 
-Third-party libraries may be required based on the CMake options selected. If the system already has the library
-installed then it will be used, otherwise CMake will retrieve the source code for the library from its official git repository and compile it in when the `MZ_FETCH_LIBS` option is enabled.
+Third-party libraries may be required based on the CMake options selected. CMake prefers an installed library,
+then source code in `third_party`, and finally retrieves the source code from its official git repository when
+`MZ_FETCH_LIBS` is enabled. `MZ_FORCE_FETCH_LIBS` skips installed libraries and fetches a dependency when no
+bundled source is available. When both fetch options are enabled, bundled sources are ignored and FetchContent
```
nmoinvaz: why is there a + here?

CMakeLists.txt
```diff
@@ -173,6 +173,27 @@ check_function_exists(readlink HAVE_READLINK)
 configure_file(mz_config.h.cmakein ${CMAKE_CURRENT_BINARY_DIR}/mz_config.h @ONLY NEWLINE_STYLE UNIX)
 list(APPEND MINIZIP_HDR ${CMAKE_CURRENT_BINARY_DIR}/mz_config.h)
 
+#
```
nmoinvaz: moving this section could be its own commit for easier reviewing

nmoinvaz (approved):

PR #1036, opened 2026-09-02
Offline compilation support

serunis:
## Summary

- Prefer installed compression libraries unless fetching is explicitly forced.
- Support pre-populated dependency sources under `<source>/third_party` when both fetch options are disabled.
- Keep fetched sources in the CMake build tree when fetching is enabled.

## Motivation

Offline and intranet builds must be able to compile without fetching from external repositories. Users can prepare the required dependency source code under `third_party` ahead of time, then configure and build minizip-ng without outbound network access.

## Why the previous code could not provide an offline build

On the base commit (`bc099de`), the local `third_party` directory was not treated as a reusable source fallback:

- ZLIB, BZIP2, LZMA, and Zstandard only entered `clone_repo()` when `MZ_FETCH_LIBS=ON`. With both fetch options disabled and no system package available, CMake did not inspect the pre-populated source directories and disabled those compression backends.
- PPMd called `clone_repo()` unconditionally. The old `FetchContent_Declare(... SOURCE_DIR <source>/third_party/ppmd)` selected the clone destination; it did not mark existing contents as already populated. CMake therefore still attempted to clone from GitHub even when `MZ_FETCH_LIBS=OFF`.

### Reproduction on the base commit

Tested on WSL2 (Debian 13, GCC 14.2.0, CMake 3.31.6) with all five dependency sources prepared under `<source>/third_party`, system package discovery disabled, both fetch options set to `OFF`, and network proxies pointed at an unreachable local endpoint:

- Configuration reported ZLIB, BZIP2, and LibLZMA as not found, then attempted to clone PPMd from GitHub three times and failed.
- Repeating the test with `MZ_PPMD=OFF` allowed configuration and compilation to finish, but ZLIB, BZIP2, LZMA, and Zstandard were all disabled despite their sources being present. CMake reported `Compression not supported due to missing libraries`.

The previous implementation therefore could not produce a full compression-enabled build from a self-contained source tree with fetching disabled: it either attempted network access for PPMd or silently disabled the other compression backends.

## Source selection by option

| `MZ_FETCH_LIBS` | `MZ_FORCE_FETCH_LIBS` | System package lookup | Dependency source location and behavior | Network fetch |
|---|---|---|---|---|
| `OFF` | `OFF` | Yes | Use the installed package when found; otherwise use pre-populated source from `<source>/third_party/<dependency>` | No. If neither source is available, the corresponding compression method is disabled. |
| `ON` | `OFF` | Yes | Use the installed package when found; otherwise populate source under `<build>/_deps/<dependency>-src` | As needed |
| `OFF` | `ON` | No | Populate source under `<build>/_deps/<dependency>-src` | As needed |
| `ON` | `ON` | No | Populate source under `<build>/_deps/<dependency>-src` | As needed |

PPMd does not have a system-package lookup path, so it uses the pre-populated `third_party` source directly in offline mode and the build-tree source in fetch-enabled modes.

## Pre-populated source directories

| Dependency | Offline/intranet source directory |
|---|---|
| ZLIB / zlib-ng | `<source>/third_party/zlib` |
| BZIP2 | `<source>/third_party/bzip2` |
| LZMA / xz | `<source>/third_party/liblzma` |
| PPMd / 7-Zip | `<source>/third_party/ppmd` |
| Zstandard | `<source>/third_party/zstd` |

## Validation

- Windows: CMake configure and build completed successfully; all 246 CTest tests passed.
- Linux (WSL2, Debian 13, GCC 14.2.0, CMake 3.31.6): with `MZ_FETCH_LIBS=OFF` and `MZ_FORCE_FETCH_LIBS=OFF`, all compression dependencies were resolved from pre-populated `<source>/third_party` sources while system package discovery was disabled and network proxies were invalid. Configure and build completed successfully, and all 246 CTest tests passed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -173,6 +173,27 @@ check_function_exists(readlink HAVE_READLINK)
 configure_file(mz_config.h.cmakein ${CMAKE_CURRENT_BINARY_DIR}/mz_config.h @ONLY NEWLINE_STYLE UNIX)
 list(APPEND MINIZIP_HDR ${CMAKE_CURRENT_BINARY_DIR}/mz_config.h)
 
+#
+# Check whether a sanitizer was requested
+#
+message(STATUS "Checking for sanitizer option")
+
+string(TOLOWER "${MZ_SANITIZER}" MZ_SANITIZER)
+if(NOT MZ_SANITIZER)
+    message(STATUS " sanitizer not enabled")
+elseif(MZ_SANITIZER STREQUAL "address")
+    add_address_sanitizer()
+elseif(MZ_SANITIZER STREQUAL "memory")
+    add_memory_sanitizer()
+elseif(MZ_SANITIZER STREQUAL "thread")
+    add_thread_sanitizer()
+elseif(MZ_SANITIZER STREQUAL "undefined")
+    add_undefined_sanitizer()
+else()
+    message(FATAL_ERROR "Unknown MZ_SANITIZER option: '${MZ_SANITIZER}'\n"
+                        "MZ_SANITIZER must be one of: 'memory', 'address', 'undefined', 'thread'")
+endif()
+
 if(MZ_LIBCOMP)
     if(APPLE)
         # Use Apple libcompression
@@ -191,6 +212,9 @@ if(MZ_LIBCOMP)
 endif()
 
 if(MZ_ZLIB)
+    set(_zlib_local_source_dir "${CMAKE_CURRENT_SOURCE_DIR}/third_party/zlib")
+    set(_zlib_source_build OFF)
+
     # Check if zlib is present
     if(NOT MZ_FORCE_FETCH_LIBS)
         if (MZ_ZLIB_FLAVOR STREQUAL "zlib-ng" OR MZ_ZLIB_FLAVOR STREQUAL "auto")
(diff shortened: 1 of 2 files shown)


nmoinvaz:
I thought we already supported this with `MZ_FORCE_FETCH_LIBS=OFF`?

serunis:
No. On Linux, the first three were likely already installed as system packages, so they didn't cause any trouble. But PPMd and zstandard reveal the problem because they aren't usually installed system-wide.

If you unplug the LAN cable, turn off Wi-Fi, and compile, PPMd will show the problem. Even after fixing PPMd, zstd will silently become unsupported, causing the "minizip -t" option to fail.

nmoinvaz:
There has to be an easier or cleaner solution right?

nmoinvaz:
Is this the minimal fix?

serunis:
I thought policy should be aligned.

The first three also work without issues only if they are installed as system packages. If they aren't installed, they cause the same problem.

nmoinvaz:
This already works on `develop` with no source changes. Keep fetching on and tell FetchContent the sources exist:

```
cmake -S . -B build -DMZ_FETCH_LIBS=ON -DMZ_FORCE_FETCH_LIBS=ON \
      -DFETCHCONTENT_FULLY_DISCONNECTED=ON
```

Or `-DFETCHCONTENT_SOURCE_DIR_<NAME>=<src>/third_party/<name>` per dependency.

Verified on a fresh build dir with proxies pointed at a dead port. All five deps came from `third_party`, no clone sub-builds ran, 246/246 tests passed.

`MZ_FETCH_LIBS=OFF` means never fetch, so deps without a system package get disabled by design. That is the cause of your reproduction.

This PR duplicates `clone_repo`, changes what `MZ_FETCH_LIBS=OFF` means, moves fetched sources out of `third_party`, and relocates the sanitizer block. I would take a README note with the recipe above, plus this, since `MZ_FORCE_FETCH_LIBS=ON` alone fetches nothing on non-Windows:

```diff
-    elseif(MZ_FETCH_LIBS)
+    elseif(MZ_FETCH_LIBS OR MZ_FORCE_FETCH_LIBS)
```

serunis:
I don't understand.  current develop below still fetches ppmd
```sh
> cmake -S . -B build -DMZ_FETCH_LIBS=OFF -DMZ_FORCE_FETCH_LIBS=OFF
-- The C compiler identification is GNU 8.5.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Using CMake version 3.26.5
-- Looking for dirent.h
-- Looking for dirent.h - found
-- Looking for sys/dirent.h
-- Looking for sys/dirent.h - not found
-- Looking for inttypes.h
-- Looking for inttypes.h - found
-- Looking for stdint.h
-- Looking for stdint.h - found
-- Looking for sys/types.h
-- Looking for sys/types.h - found
-- Looking for stddef.h
-- Looking for stddef.h - found
-- Check size of DIR*
-- Check size of DIR* - done
-- Check size of off64_t
-- Check size of off64_t - failed
-- Looking for fseeko
-- Looking for fseeko - found
-- Looking for symlink
-- Looking for symlink - found
-- Looking for readlink
-- Looking for readlink - found
-- Using ZLIB 1.2.11
-- Using BZIP2 1.0.6
-- Using LZMA 5.2.4
-- Using PPMD
-- Fetching ppmd https://github.com/ip7z/7zip 26.02
[ 11%] Creating directories for 'ppmd-populate'
[ 22%] Performing download step (git clone) for 'ppmd-populate'
Cloning into 'ppmd'...
fatal: unable to access 'https://github.com/ip7z/7zip/': Could not resolve host: github.com
Cloning into 'ppmd'...
fatal: unable to access 'https://github.com/ip7z/7zip/': Could not resolve host: github.com
Cloning into 'ppmd'...
fatal: unable to access 'https://github.com/ip7z/7zip/': Could not resolve host: github.com
-- Had to git clone more than once: 3 times.

```

serunis:
understood,

1. copy ppmd, zstd to third_party/ppmd third_party/zstd
2. cmake -S . -B build -DMZ_BUILD_TESTS=ON -DMZ_FETCH_LIBS=OFF -DMZ_FORCE_FETCH_LIBS=OFF -DFETCHCONTENT_FULLY_DISCONNECTED=ON
3. cmake --build build

oh, fetch does not mean downloading.. fetch sounds like downloading

"-DFETCHCONTENT_FULLY_DISCONNECTED=ON" is the offline option I wanted, thank you

PR #1046, opened 2026-09-06
Fix heap buffer overflow on encrypted entries with an undersized compressed size

semvis123:
`mz_zip_entry_open_int` sets the crypt stream's `max_total_in` to the
`compressed_size` from the central directory, and `mz_stream_wzaes_read` and
`mz_stream_pkcrypt_read` clamp each read to what is left of it. An encrypted
entry also has to hold its own crypto header and footer. The header is counted
into `total_in` when the stream is opened, and `mz_stream_wzaes_read` subtracts
`MZ_AES_FOOTER_SIZE` on top of that, so an entry declaring a compressed size
below that overhead drives the clamp negative.

The negative `int32_t` is passed to `mz_stream_read` and reaches
`mz_stream_os_read`, where it sign extends into a huge `size_t` in `fread` and
copies the archive into the caller's buffer until EOF. The length and the
contents of the overflow both come from the archive, so a crafted entry gives a
controlled heap buffer overflow (CWE-195 signed to unsigned conversion leading
to CWE-787 out of bounds write). `mz_zip_entry_read` already returns 0 for
`compressed_size == 0`, so the range that triggers it is
`0 < compressed_size < header + footer`.

`mz_zip_reader_open_file` is not affected: `mz_stream_buffered_read` returns 0
for a negative size and absorbs it, so the `minizip` CLI extracts a crafted
archive without complaint. Callers passing their own `mz_stream_os` to
`mz_zip_reader_open` or `mz_zip_open` are the ones that overflow.

`mz_zip_entry_open_int` now returns `MZ_FORMAT_ERROR` for an entry whose
compressed size cannot hold its crypto header and footer, and `mz_stream_read`
rejects a negative size so a miscomputed clamp elsewhere cannot reach `fread`
or `ReadFile`.

### DIFF
--- a/mz_strm.c
+++ b/mz_strm.c
@@ -33,10 +33,7 @@ int32_t mz_stream_is_open(void *stream) {
 
 int32_t mz_stream_read(void *stream, void *buf, int32_t size) {
     mz_stream *strm = (mz_stream *)stream;
-    if (!strm || !strm->vtbl || !strm->vtbl->read)
-        return MZ_PARAM_ERROR;
-    /* A negative size sign-extends to a huge length in fread/ReadFile and overflows the caller's buffer */
-    if (size < 0)
+    if (!strm || !strm->vtbl || !strm->vtbl->read || size < 0)
         return MZ_PARAM_ERROR;
     if (mz_stream_is_open(strm) != MZ_OK)
         return MZ_STREAM_ERROR;
--- a/mz_strm.c
+++ b/mz_strm.c
@@ -35,6 +35,9 @@ int32_t mz_stream_read(void *stream, void *buf, int32_t size) {
     mz_stream *strm = (mz_stream *)stream;
     if (!strm || !strm->vtbl || !strm->vtbl->read)
         return MZ_PARAM_ERROR;
+    /* A negative size sign-extends to a huge length in fread/ReadFile and overflows the caller's buffer */
+    if (size < 0)
+        return MZ_PARAM_ERROR;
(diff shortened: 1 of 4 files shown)


mz_strm.c
```diff
@@ -35,6 +35,9 @@ int32_t mz_stream_read(void *stream, void *buf, int32_t size) {
     mz_stream *strm = (mz_stream *)stream;
     if (!strm || !strm->vtbl || !strm->vtbl->read)
         return MZ_PARAM_ERROR;
+    /* A negative size sign-extends to a huge length in fread/ReadFile and overflows the caller's buffer */
+    if (size < 0)
```
nmoinvaz: merge this into previous if statement, get rid of comment.

mz_zip.c
```diff
@@ -1855,7 +1855,11 @@ static int32_t mz_zip_entry_open_int(void *handle, uint8_t raw, int16_t compress
                 if (mz_stream_get_prop_int64(zip->crypt_stream, MZ_STREAM_PROP_FOOTER_SIZE, &footer_size) == MZ_OK)
                     max_total_in -= footer_size;
 
-                mz_stream_set_prop_int64(zip->compress_stream, MZ_STREAM_PROP_TOTAL_IN_MAX, max_total_in);
+                /* Reject an entry whose compressed size cannot hold the encryption header and footer */
```
nmoinvaz: no need for comment. its obvious.

nmoinvaz (changes requested):

nmoinvaz (approved):

PR #1064, opened 2026-09-13
Update central directory bounds when replacing its stream

### LINKED ISSUE #1063: `mz_zip_set_cd_stream` keeps stale central directory bounds
Hello! I found this while using `mz_zip_set_cd_stream` to provide a separately decrypted central directory to minizip-ng.

The current implementation replaces `cd_stream` and `cd_start_pos`, but keeps the previous `cd_size`:

https://github.com/zlib-ng/minizip-ng/blob/32455ac154f63f37e0ec9bfa7d9345cddeb9e649/mz_zip.c#L1632-L1641

This can lead to inconsistent behavior when the replacement central directory has a different size:

- sequential iteration over the replacement directory works.
- `mz_zip_goto_entry()` can reject a valid entry offset with `MZ_PARAM_ERROR` because it is still checked against the old `cd_size`.

Given that `cd_stream` is documented as the stream used for reading the central directory, the most natural fix seems to be updating `cd_size` when the stream is replaced:

```c
cd_size = stream_end - cd_start_pos;
```

**I'm about to send a PR with a fix, but I'd like to align on this API detail first so the fix matches the intended contract:**

Both current internal callers, directory recovery and CDCD extraction, supply dedicated directory streams. If that is the intended contract, this calculation seems appropriate and would also prevent those paths from retaining stale bounds that can cause `mz_zip_goto_entry()` to reject valid offsets.

However, if the setter also accepts a complete ZIP stream with `cd_start_pos` pointing to its directory, the calculation would overestimate the size. The directory is followed by EOCD record and, where applicable, ZIP64 end records, as shown in [PKWARE APPNOTE 4.3.6](https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT).

I traced the internal `cd_start_pos` field to its [original introduction](https://github.com/zlib-ng/minizip-ng/commit/cda3600384a624ba8e2d75546a130d7e80fc9aca), which supported **reading the directory** from either the original archive or a separate memory stream.

I could not find why this setter exposed this parameter, so that history does not establish whether trailing data is allowed in a replacement stream.

So, I'd like to clarify this before submitting a PR:

- Should the replacement directory extend from `cd_start_pos` to the end of `cd_stream`, or should passing a complete ZIP stream also be supported?
- If trailing data is allowed, would you prefer adding an explicit size parameter to `mz_zip_set_cd_stream`, which would be a breaking change to the public API, or introducing a separate size setter to preserve compatibility with existing callers?

igooor-bb:
Hello! This PR addresses the stale central directory bounds discussed in #1063.

### Problem

`mz_zip_set_cd_stream` replaces the stream and its start position but keeps the previous `cd_size`. With a larger replacement directory, sequential iteration can succeed while `mz_zip_goto_entry()` rejects valid offsets with `MZ_PARAM_ERROR`.

### Fix

Update `cd_size` to `stream_end - cd_start_pos`, treating the replacement directory as extending to the stream's end.

The setter validates the bounds, restores the stream position and updates the ZIP fields only after success. It uses `tell -> seek to end -> tell -> seek back` because the stream API has no general length property. Adding one would require defining its semantics and implementing support across stream types, making this a broader change.

The original setter did not move the stream position, so the fix saves and restores it to avoid introducing an unexpected side effect for callers.

### Validation

I kept two regression tests for replacing the directory at zero and nonzero offsets. They check that valid entries remain accessible, offsets beyond the new directory are rejected, and the stream position is preserved. The tests focus on the changed behavior rather than every possible input or error path.

All 69 stock CTest checks passed on macOS.

Closes #1063


nmoinvaz:
I don't want the unit test at all. I don't see it as having enough value. And it looks complicated, doesn't use a fixture, introduces a new test file with a name that I'm unsure about, etc. I just don't want to deal with going through all that.

igooor-bb:
Fair enough, I've removed the test

nmoinvaz:
Needs rebase thanks.

PR #1066, opened 2026-09-13
compat: include the zlib API the library was built against

biodranik:
`compat/unzip.h` and `compat/zip.h` preferred `zlib-ng.h` whenever it was on the include path. On systems that install zlib-ng's native header next to zlib, such as Arch with the `zlib-ng` package, consumers of a zlib-compat build got both headers and failed with conflicting declarations.

The headers now include `<zlib.h>` when `ZLIB_COMPAT` is defined, and the build exports `ZLIB_COMPAT` through CMake and pkg-config when `MZ_COMPAT` is enabled. A separate commit guards the `__has_include` probe with `defined(__has_include)`, so compilers without it, such as GCC 4.8, can preprocess the headers.

Hit it in https://github.com/organicmaps/organicmaps/issues/13554 and worked around it in https://github.com/organicmaps/organicmaps/pull/13562


compat/unzip.h
```diff
@@ -19,14 +19,27 @@ extern "C" {
...
-#  if __has_include(<zlib-ng.h>)
+/* Pull in the same zlib flavour minizip-ng was built against. MZ_ZLIB_COMPAT is
+   supplied by the build when the library was compiled for a zlib-compatible
+   API; probing the include path is only a fallback for builds that cannot pass
+   the define, and it picks the wrong header on systems where zlib-ng's native
+   headers are installed next to zlib. */
+#if !defined(_ZLIB_H) && !defined(ZLIB_H) && !defined(ZLIB_H_) && !defined(ZNGLIB_H_)
+#  if defined(MZ_ZLIB_COMPAT) || defined(ZLIB_COMPAT)
+#    include <zlib.h>
+#  elif __has_include(<zlib-ng.h>)
```
nmoinvaz: don't you have to check if `__has_include` even exists?

compat/unzip.h
```diff
@@ -19,14 +19,27 @@ extern "C" {
...
 
-#if !defined(_ZLIB_H) && !defined(ZLIB_H) && !defined(ZLIB_H_)
-#  if __has_include(<zlib-ng.h>)
+/* Pull in the same zlib flavour minizip-ng was built against. MZ_ZLIB_COMPAT is
+   supplied by the build when the library was compiled for a zlib-compatible
+   API; probing the include path is only a fallback for builds that cannot pass
+   the define, and it picks the wrong header on systems where zlib-ng's native
+   headers are installed next to zlib. */
+#if !defined(_ZLIB_H) && !defined(ZLIB_H) && !defined(ZLIB_H_) && !defined(ZNGLIB_H_)
+#  if defined(MZ_ZLIB_COMPAT) || defined(ZLIB_COMPAT)
```
nmoinvaz: why do you need a `MZ_ZLIB_COMPAT`, we already define `ZLIB_COMPAT` in CMake so why not use it?

compat/unzip.h
```diff
@@ -19,14 +19,27 @@ extern "C" {
 
 #include <stdint.h>
 
-#if !defined(_ZLIB_H) && !defined(ZLIB_H) && !defined(ZLIB_H_)
-#  if __has_include(<zlib-ng.h>)
+/* Pull in the same zlib flavour minizip-ng was built against. MZ_ZLIB_COMPAT is
```
nmoinvaz: This comment is overkill

nmoinvaz (commented):
The comments are books. I don't have time to real them.

compat/unzip.h
biodranik: Removed. I also dropped the `#error` block, which rejected files that include `zlib-ng.h` on purpose.

compat/unzip.h
biodranik: Agreed. `ZLIB_COMPAT` was private to the target, so consumers never saw it. It is now exported when `MZ_COMPAT` is on, and `MZ_ZLIB_COMPAT` is gone.

compat/unzip.h
biodranik: Yes. It was unguarded before this PR too, and GCC 4.8 fails on it. The probe is now nested under `defined(__has_include)` like in `mz.h`, in a separate commit.

